PR #30, opened 2016-07-21
Update node vertical test to use pause pods

mffiedler:
The node vertical test currently uses the hello-openshift pod.   While smaller than some pods, it is still too large for a test which has a goal of solely measuring pod overhead.   The gcr.io/google_containers/pause-amd64:3.0  pd is pretty much the smallest possible shim pod available.  Update the nodeVertical test case to use this pod.

Also, for 3.3, update the number of pods created across 2 nodes to 500.


jeremyeder:
Cool, thanks!

PR #49, opened 2016-08-01
docker-py is required by ansible to use docker module

jeremyeder:
For some reason ansible doesn't try to pull down docker-py.  It's required in order to use ansible's docker module.


jeremyeder:
https://github.com/ansible/ansible/issues/16906

PR #62, opened 2016-08-29
Remove clean support from cluster-loader

hroyrh:
It was not being used and there were no further plans to use it so after a discussion we decided to remove it. 
@mffiedler can you take a look at this.
I was thinking that maybe we can add the delete by label feature within cluster-loader itself, I can update the PR with that if we want it. What are you thoughts on it @timothysc @jeremyeder


timothysc:
We should probably allow an option for the input label used on bulk deletes, but that can be done separately.

PR #67, opened 2016-09-09
AWS compatible initial commit

sjug:
$title


image_provisioner/cloud-init/user-data
```diff
@@ -9,3 +9,5 @@ user: root
 password: 100yard-
```
jeremyeder: nope

image_provisioner/inventory/hosts
```diff
@@ -0,0 +1,55 @@
...
+# collectd install
+config_type=ose
+collectd_interval=10
+collectd_status=started
+graphite_host=
+graphite_prefix=ose
+# clone repos
+github_token=
+# pbench config
+image_tag=2016-09-09
```
jeremyeder: I tried to figure out a way to get this to be auto-calculated at runtime and the feedback I got was:

You could assemble the timestamp using facts or using a lookup, same goes for expanding the home directory.

http://docs.ansible.com/ansible/playbooks_variables.html#information-discovered-from-systems-facts
http://docs.ansible.com/ansible/playbooks_lookups.html

image_provisioner/playbooks/build_ami.yaml
```diff
@@ -0,0 +1,77 @@
...
+      assign_public_ip: yes
+      region: "{{ region }}"
+      key_name: "{{ key_name }}"
+      group_id: "{{ group_id }}"
+      instance_type: "{{ instance_type }}"
+      vpc_subnet_id: "{{ vpc_subnet_id }}"
+      image: "{{ base_image }}"
+      volumes:
+        - device_name: /dev/sdb
+          volume_type: io1
```
jeremyeder: I think we decided to make this gp2.  It's important to get right because io1 was very expensive.  @mffiedler @ekuric what was the conclusion?

image_provisioner/playbooks/build_ami.yaml
```diff
@@ -0,0 +1,77 @@
...
+          volume_size: 30
```
jeremyeder: And @mffiedler @ekuric was 30G ok ?

image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
```diff
@@ -26,3 +31,4 @@
...
...ckages/setools/3.3.8/1.1.el7/x86_64/setools-libs-3.3.8-1.1.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/libsepol/2.5/4.el7/x86_64/libsepol-2.5-4.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/libsemanage/2.5/3.el7/x86_64/libsemanage-python-2.5-3.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/libselinux/2.5/4.el7/x86_64/libselinux-2.5-4.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/libselinux/2.5/4.el7/x86_64/libselinux-utils-2.5-4.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/systemd/219/24.el7/x86_64/systemd-219-24.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/systemd/219/24.el7/x86_64/systemd-libs-219-24.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/systemd/219/24.el7/x86_64/systemd-sysv-219-24.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/systemd/219/24.el7/x86_64/libgudev1-219-24.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/libselinux/2.5/4.el7/x86_64/libselinux-python-2.5-4.el7.x86_64.rpm http://{{ rhel_repo_hostname }}/brewroot/packages/iptables/1.4.21/17.el7/x86_64/iptables-services-1.4.21-17.el7.x86_64.rpm
```
jeremyeder: For some context, these overrides are only necessary because of current point of the development cycle we're in in RHEL-land.  We eventually should get away from this because static versions are a bear to manage...basically every time we build an image we're going to have to update this step (and I've been doing this manually).

image_provisioner/playbooks/roles/repo-install/files/rhel72.repo
```diff
@@ -0,0 +1,29 @@
+[rhel-72]
+name=Prerelease version of RHEL7.2
+baseurl=https://mirror.openshift.com/enterprise/rhel/rhel7next/os
```
jeremyeder: hostname should be taken from the inventory file.

image_provisioner/playbooks/build_ami.yaml
mffiedler: The default needs to be gp2.   People who know that they need io1 can change it.   We also need to make sure that the option deleteOnTermination is set to true for /dev/sdb in the AMI.

image_provisioner/playbooks/build_ami.yaml
mffiedler: We override it in SVT in our AMIs.   When doing a lot of builds for different apps, we use 80G.   But, I think 30G is a good default and if folks need more, they can override on instance creation.    Going too big is wasteful because you can't go lower when creating the instance.

jeremyeder: We'd like, if possible, that this work be used by both teams so that there we're both using the same AMIs.  Hopefully account for whatever unique case you might have (disk space...any others?).  Is this reasonable?

image_provisioner/playbooks/roles/repo-install/files/rhel72.repo
sjug: I might be able to pull that repo file from the aos-ansible repo as well... Once I get that in before git

image_provisioner/cloud-init/user-data
sjug: That was in the file originally, but I'll get rid of it.

image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
sjug: Indeed, this is gross and hopefully very temporary.

image_provisioner/inventory/hosts
sjug: @jeremyeder, expanding the home directory for what info?

sjug: Replaced image tag with ansible_date_time.date

sjug:
Let me know if I should make any other changes :+1:

mffiedler:
I will test this week with this PR + the aos-ansible change.

sjug:
Thanks @mffiedler, I have a new commit on the way to fix the ssh keys. Let me know how it goes!

sjug:
The docker version override task in the docker-config role fails silently on AWS, since the repo is internal doesn't pull the right versions down.

image_provisioner/inventory/hosts
```diff
@@ -0,0 +1,55 @@
...
+openshift_version=3.3.0.30-1
+
+# qcow specific
+#ansible_ssh_private_key_file = $HOME/.ssh/id_rsa_perf
+ansible_ssh_private_key_file = 
+base_image_url=http://download.url/rhel-guest-image-7.2-20160302.0.x86_64.qcow2
+base_image_filename=rhel-guest-image-7.2-20160302.0.x86_64.qcow2
+content_mirror=false
+# amount of GB to extend root partition by.  default size is 10G so extend that by an additional # gb:
+extend_disk_gb=15
```
mffiedler: When extend_disk is specified, the root_partition var is apparently needed.   Running the sample from svt-private, hit an error that root_partition was not defined.

image_provisioner/inventory/hosts
```diff
@@ -0,0 +1,55 @@
...
+# clone repos
+github_token=
+# pbench config
+pbench_internal_hostname=pbench\..*com$
+pbench_alternate_hostname=
+# docker config
+# sda2 for openstack, sdb/xvdb for ec2
+config_repo=
+docker_login=no
+docker_storage_device=/dev/xvdb
```
mffiedler: Running the setup-image.yaml using this inventory - it could not find xvdb on the resize second partition step.   If I use vda2 it works.   Am I doing something wrong with the base image?  or should be running a different inventory for setup-image.yaml?

mffiedler:
Ignore deleted comments on partition names.   Bad starting assumption on my part.     We need README updates with AWS instructions, including prereqs (e.g. boto).

sjug:
@mffiedler, I already had any prereqs installed before I made the script it seems. I would like to make the ansible playbook self-contained so that it checks/installs prereqs locally on execution.

Was it missing python2-boto or python3-botocore? Anything else?

mffiedler:
Suggestion:  tag the instance in AWS in case there is a failure prior to playbook completion.   Makes the instance more identifiable.

@sjug  My test made it to the "terminate temporary instance" task and failed with:

TASK [terminate temporary instance] ********************************************
Wednesday 14 September 2016  09:12:22 -0400 (0:00:01.689)       0:23:49.964 *** 
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "No handler was ready to authenticate. 1 handlers were checked. ['HmacAuthV4Handler'] Check your credentials"}

The AMI was not created.   Any ideas?

sjug:
@mffiedler: I changed the tag from role to Name so it'll be visible by default on AWS console.

The terminate task entry is saying that it can't find the AWS credentials, which is strange and I don't have that issue. We can force add them to that task as well?

jeremyeder:
Ahh this is a really good point...the Name tag can be used to reconcile billing.  Can you add another tag called group= and then wire that into the inventory?  Possible groups would be QE or Perf for now.  Let me know if you prefer this be a separate github issue.

sjug:
@jeremyeder: yep not a problem at all, I'll push that up now

sjug:
@mffiedler Are you still seeing that issue with the terminate step? I did a fresh test today and I can't reproduce the credential error.

image_provisioner/inventory/inventory
```diff
@@ -1,36 +0,0 @@
...
-# this is the version of the images.  never has the trailing -N like the RPMs might
-image_version=v3.3.0.11
-qe_repo_image_type=ose
-rhel_repo_hostname=
-pbench_internal_hostname=
-pbench_alternate_hostname=
-root_password=
-
-[all]
-# 127.0.0.1
```
mffiedler: We shouldn't delete this file entirely since it is a good OpenStack example.   I suggest having inventory.OS and inventory.AWS.   We can add inventory.GCE etc later.

sjug: No problem I can refactor the initial inventory a bit and retain & rename it to suit.

PR #72, opened 2016-09-22
os_kickstart must be first role

mffiedler:
If os_kickstart is not first, nobody will be happy.


mffiedler:
Need to test this better.   New PR coming shortly

PR #73, opened 2016-09-22
Small fixes for OpenStack provisioning

mffiedler:
1. Fix order of roles in provision_gold_images.yaml.  repo-install and os-kickstart have to be first for everyone to be happy.
2. typo in iscsi-initiator-utils package name
3. no need to override docker version right now.


mffiedler:
Remove cncf config

PR #93, opened 2016-11-15
RHEL 7.3 support and consolidated pbench repos

mffiedler:


mffiedler:
Internal CDN site does not have final RHEL 7.3 or Extras bits.   Switched qcow2 provisioner to external=true for now so that it uses the mirror for latest RHEL 7.3 bits.

Consolidated pbench repos in single repo file.

Committed a sample inventory for 7.3 to svt-private.

PR #111, opened 2017-01-04
Updates to network tests and app performance tests

schituku:
The following updates are made to network tests:
   - network tests for services are now able to run with multiple ports used by uperf
   - the pod and service tests will run for both stream and rr for UDP
The following updates are made to app performance tests
   - The ability to draw graphs for response times as the user count varies is added
   - Removed typos from the graphs legends


mffiedler:
@schituku  Prior to merge we need to break this into commits for each of the functional areas there are changes (networking, sockets, app performance, etc).   I can help with that.

openshift_scalability/content/quickstarts/eap/eap64-mysql-pv.json
```diff
@@ -675,7 +675,6 @@
                                         "value": "${JGROUPS_CLUSTER_PASSWORD}"
                                     }
                                 ]
-				}
```
mffiedler: This change and the other similar ones in this commit don't seem to be necessary.   When I look at the files on master they do not have this extraneous close brace.

schituku:
@mffiedler  Broken this PR into three different PRs with related changes clubbed together. I omitted the commits that made changes to the resource limits in Kitchen Sink Tests.

mffiedler:
Closing this PR in  favor of the more targeted PRs

PR #114, opened 2017-01-05
Updates to Network Synthetic tests

schituku:


mffiedler:
Tested and verified the tests now run ok.   The approach of having to add all of the application ports in the template is pretty ugly, but since we are considering a refactor of the networking tests, merging this so that the tests can continue to be run as-is and results compared with previous releases which used this same methodology.

PR #124, opened 2017-01-18
Add install external dockernext repo to image_provisioner/playbooks/r…

wabouhamad:
Updated playbook task to install external dockernext repo in file image_provisioner/playbooks/roles/repo-install/tasks/main.yaml


jeremyeder:
I think this belongs in aos-ansible, no?  (that's where we handle all the yum repos for product content)

wabouhamad:
@jeremyeder:  several yum repos including rhel7next are installed in that one role "repo-install" task.  I added the dockernext repo to that same task.

Issue #135, opened 2017-01-26
Networking: Synthetic tests use deprecated oc process options


schituku:
As of 3.5 oc process will not use -v option and suggest usage of -p. The network synthetic tests should be fixed to use oc process -p instead of oc process -v

PR #136, opened 2017-01-26
Fix deprecated usage of -v param for oc process command in 3.5 version.

schituku:
(cherry picked from commit 91c9fca70e1774445f14501e0a3c72971a4da4ec)

fix for #135

### DIFF
--- a/networking/synthetic/pod-ip-test-setup.yaml
+++ b/networking/synthetic/pod-ip-test-setup.yaml
@@ -62,8 +62,8 @@
   - name: Create sender pods
     shell: >
       oc process
-      -v REGION={{ sender_region }}
-      -v ROLE=sender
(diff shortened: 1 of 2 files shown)


vikaschoudhary16:
This change has broken backward compatibility. Will it make sense if a check is added to verify version and behave accordingly?

mffiedler:
@vikaschoudhary16  I opened #147 for this and meant to fix it by now.   I will make it a priority.   Sorry for the inconvenience.

mffiedler:
@vikaschoudhary16 @schituku #147 actually fixes this for cluster-loader, not the network tests.

@schituku Can you please make a change so that the network tests work with 3.5 as well as with 3.4 and earlier.   I opened #153 for this.

PR #137, opened 2017-01-27
Added role seal-image to playbook build_ami.yaml and updated seal-ima…

wabouhamad:
…ge task to only have ym clean all step.  This is the last yum clean all before bundling ami

### DIFF
--- a/image_provisioner/playbooks/roles/seal-image/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/seal-image/tasks/main.yaml
@@ -1,7 +1,8 @@
 ---
-- name: clean yum cache
+- name: clean yum cache final time before bundle ami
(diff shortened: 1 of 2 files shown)


image_provisioner/playbooks/roles/seal-image/tasks/main.yaml
```diff
@@ -1,7 +1,8 @@
 ---
-- name: clean yum cache
+- name: clean yum cache final time before bundle ami
```
mffiedler: nit: Seal image is used for qcow2 also, not just AMIs.   Comment a little misleading.

wabouhamad: Comment updated

PR #154, opened 2017-02-16
dynamic persistent storage update for applications templates

ekuric:
dynamic persistant storage support for applications templates 
check help.md for short guidance how to use them.
these templates does not create storageclasses but only use them to allocate storage, that means, prior using these templates ensure that Openshift is properly configured from storage and storageclasses point of view.

### DIFF
--- a/openshift_scalability/content/quickstarts/help.md
+++ b/openshift_scalability/content/quickstarts/help.md
@@ -1,0 +1,106 @@
+### Usage
+
+Templates in this directory can be used with cluster loader with
+
+```
+# python cluster-loader.py -f appdeloy.yaml
+```
+where in `appdeloy.yaml` we can specify `storageclass` which will be used by
+application pods to allocate storage
+
+Beside `STORAGE_CLASS` parameter, we can specify also other parameters related storage
+as `ACCESS_MODES` where access mode is one from https://kubernetes.io/docs/user-guide/persistent-volumes/#access-modes
+
+`VOLUME_CAPACITY` parameter allows to specify size of volume which will be mounted
+inside pod
+An example of appdeploy.yaml is showed below.
+
+Below template will
+
+- create one project
+- use templates specified by `file` line. Currently these templates are tested, it is possible to add more templates, but not tested more at this time.
+-  applications starting based on specific template, will use storage class
+specified with variable `STORAGE_CLASS` to allocated persistent volume claims(PVC) for applications. The storageclass must exist prior trying to use it. For instructions how to configure storage class, check OCP documentation
+- `ACCESS_MODES` set to be `ReadWriteOnce`
+- VOLUME_CAPACITY set to 5Gi
+
+Mount point where PVC is mounted inside application pods is managed by applications
+and we cannot add it at time as parameter.
+
+In below template, it is possible to use different storage classes for different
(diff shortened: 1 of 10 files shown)


mffiedler:
@schituku The PR we discussed.  Please review when you have a chance.

openshift_scalability/content/quickstarts/help.md
```diff
@@ -0,0 +1,106 @@
+### Usage
+
+Tempaltes in this directory can be used with cluster loader with
```
mffiedler: Typo:  Templates

openshift_scalability/content/quickstarts/help.md
```diff
@@ -0,0 +1,106 @@
...
+as `ACCESS_MODES` where access mode is one from https://kubernetes.io/docs/user-guide/persistent-volumes/#access-modes
+
+`VOLUME_CAPACITY` parameter allows to specify size of volume which will be mounted
+inside pod
+An example of appdeploy.yaml is showed below.
+
+Below template will
+
+- create one project
+- use templates specified by `file` line. Currently these templates are tested, it is possible to add more tempplates, but not tested more at this time.
```
mffiedler: Typo:  templates

openshift_scalability/content/quickstarts/help.md
ekuric: @mffiedler  corrected , thank you for checking

ekuric: @mffiedler corrected

jeremyeder:
Any reason this shouldn't be rebased and merged?

mffiedler:
Will test this week.

mffiedler:
@ekuric  When I run cluster loader with master-vert-pv.yaml, the pvc doesn't bind.   I get events with:

2s         1m          6         mysql                           PersistentVolumeClaim                                 Warning   ProvisioningFailed   persistentvolume-controller                            storageclass.storage.k8s.io "foo" not found

Is this a 3.6 vs 3.5 difference in the storageclass names?

mffiedler:
@ekuric my understanding is that the storage class should be "" to pick up the default.   True?   Should I just merge this and fix it afterwards myself?

ekuric:
@mffiedler storageclass name has to be defined 
eg: 

parameters:
    - STORAGE_CLASS: "elclass"

and it must be present on system , in above case "elclass". It expects to get specific storage class - not default.

mffiedler:
@ekuric thanks.  guess I am confused on why I was seeing the error for "foo".  It's been a while, so I will re-test.   Would like to get this merged or closed once and for all.

mffiedler:
@ekuric  Results from further testing.   If I take the storageclass metadata notation out  of the quickstart templates altogether, it will successfully use the cluster default storageclass for the new pvc.     Not sure what we should do here.   Options:

1. remove STORAGE_CLASS as a parameter and use the default storageclass.   That's not good because it doesn't allow testing of other storageclasses.  Plus having a default storageclass is not required.
2. change the default from foo to something that might work for some people (like gp2)  - or update the yaml files in the config dir to specify such a default.

Opinion?

mffiedler:
Closing in favor of work on golang cluster loader

Issue #153, opened 2017-02-16
Network tests broken on 3.4


mffiedler:
This is due to #136 and should be fixed similar to #147

schituku:
PR #162 to make the network tests backward comaptible

mffiedler:
resolved

PR #162, opened 2017-02-24
Make network test backward compatible with 3.4 and earlier versions

schituku:
Added a new -v/--version option to the test which will decide the appropriate options for the versions


schituku:
fixes #153

networking/synthetic/config.yaml
```diff
@@ -1,4 +1,5 @@
 ---
 environment:
-  master: master.example.com
-  nodes: node1.example.com node2.example.com
\ No newline at end of file
+  master: ip-172-31-22-251.us-west-2.compute.internal
+  nodes: ip-172-31-13-12.us-west-2.compute.internal ip-172-31-33-87.us-west-2.compute.internal
+  version: 3.5
```
mffiedler: Please remove specific AWS addresses

networking/synthetic/network-test.py
```diff
@@ -71,6 +72,12 @@ def parse_args():
     parser.add_argument('test_type',
                         choices = ['podIP', 'svcIP'])
     
+    parser.add_argument('-v',
+                        '--version',
+                        required = True,
```
mffiedler: Might be nicer to default to 3.5 mode

networking/synthetic/config.yaml
schituku: @mffiedler  I realized this after the commit so corrected in the second commit with the comment - "new config file"

networking/synthetic/network-test.py
schituku: @mffiedler made the 3.5 as default

schituku:
@mffiedler closed the review comments and tested it. I think it is good to go for a merge.

PR #164, opened 2017-02-28
Application scale up/scale down automation

mffiedler:
Ansible playbook to 

- run cluster loader and create a "no resource limit" version of the quickstarts.   This is needed to scale to 200+ replicas
- make sure all quickstarts are available
- run scale_test.py to scale applications up/down and record the amount of time required.


mffiedler:
I'll squash before merge

PR #166, opened 2017-03-02
pvc template to create desired number of PVCs

ekuric:
Example of : 

- pvc template to create desired number of PVCs and bound them to PVs

- it uses storageclass and dynamic provisioning when creating PVC

- This will help when we want *only* to check how fast PV/PVC are created/allocated when different storageclasses ( originating from different storage backends ) are used 


jeremyeder:
Ready for merge?

ekuric:
@jeremyeder from my side it is. This template will help to create desired number of PVCs on top of storage class ( just PV and PVC without pods ) using cluster loader.

PR #171, opened 2017-03-08
README for router-arp_stress config

mffiedler:
Proposed README for the router-arp_stress.yaml configuration


mffiedler:
Need to remove extraneous commits

PR #177, opened 2017-03-18
reduce forks to what we actually use

jeremyeder:


mffiedler:
The actual max supported value for forks is 20.  Starting in 3.3, at around 50 forks, you'll start seeing ssh failures unless you bump MaxSessions in sshd_config on the masters.   See https://bugzilla.redhat.com/show_bug.cgi?id=1382492 

/cc: @detiber @brenton @sdodson

jeremyeder:
Hmm, 100 was used by @vikaslaad

jeremyeder:
@mffiedler thanks -- updated this PR to set forks = 20

PR #187, opened 2017-03-29
STAC-N1 test harness containerization

vikaschoudhary16:


vikaschoudhary16:
cc @jeremyeder @psuriset

jeremyeder:
LGTM for a first phase.  We're not handing these tests over but would still like this merged @mffiedler

PR #190, opened 2017-04-03
Baremetal playbook + enhancements

sjug:
Many commits for the time being so reviewers can decide what exactly to push to the repo.

In addition to net new roles, there are also some enhancements to the existing ones.

@mffiedler @jeremyeder


image_provisioner/playbooks/roles/create-local-repo/vars/main.yaml
```diff
@@ -0,0 +1,18 @@
+---
+rpms_to_install:
+  - createrepo
+  - nginx
```
jeremyeder: what repo does this rpm come from?

image_provisioner/playbooks/roles/dns-client-deploy/tasks/main.yaml
```diff
@@ -0,0 +1,90 @@
+---
+- name: set root password
+  user:
+    name: root
+    password: $6$zStNEJeCGkqDDaHT$VvaOFoBJB/0Yhf9Fd76mlHFFbLvMRKNAacT42de9Dk7NCPl7S4IEd9yMmAQW4hyrrbvvrrLjv/lNj6H4rChor0
```
jeremyeder: should this hash be here?

jeremyeder: do we want to merge this dns thing or move to the e2e team's solution?

image_provisioner/playbooks/roles/dns-client-deploy/tasks/main.yaml
```diff
@@ -0,0 +1,90 @@
...
+    regexp: '^nameserver {{ existing_dns_ip }}'
+    state: absent
+
+- name: get rid of resolv.conf.save
+  file:
+    path: /etc/resolv.conf.save
+    state: absent
+
+- name: get ip address
+  shell: ip addr show bond0.7 | grep -m1 -Po 'inet \K[\d.]+'
```
jeremyeder: hang on, there's a specific interface name here.

image_provisioner/playbooks/roles/dns-client-deploy/vars/main.yaml
```diff
@@ -0,0 +1,4 @@
+---
+dns_domain: "cluster.local"
+kube_svc_ip: "10.192.0.1"
+existing_dns_ip: "10.1.16.4"
```
jeremyeder: is it worth somehow parameterizing these ips

image_provisioner/playbooks/roles/dns-server-deploy/templates/dhcpd.conf.j2
```diff
@@ -0,0 +1,15 @@
+option domain-name "{{ host }}";
```
jeremyeder: this is untested/unused, right?  is it worth merging?

image_provisioner/playbooks/roles/dns-server-deploy/vars/main.yaml
```diff
@@ -0,0 +1,10 @@
+---
+host: "ocp.net"
+oct1: 10
+oct2: 2
+dns_ip: "10.2.1.3"
```
jeremyeder: variable?

image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
```diff
@@ -5,29 +5,39 @@
     state: enforcing
```
jeremyeder: the default driver remains devicemapper unless the user specifies overlay in their inventory, correct?

image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
```diff
@@ -38,22 +48,26 @@
...
  
    - name: waiting for server to come back
-     local_action: wait_for host={{ inventory_hostname }} state=started delay=15 timeout=300
+     local_action: wait_for
+     args:
+       host: "{{ inventory_hostname }}"
+       port: 22
+       state: started
+       delay: 15
+       timeout: 300
```
jeremyeder: 300 is dangerously close to what it takes for a bare metal machine to boot up; any reason not to double this just for good measure?

image_provisioner/playbooks/roles/docker-save/files/docker_save.sh
```diff
@@ -0,0 +1,7 @@
+#/bin/bash
```
jeremyeder: there is an atomic save command...can you please look at that to see if it can replace this script?

image_provisioner/playbooks/roles/docker-spread-images/tasks/main.yaml
```diff
@@ -0,0 +1,24 @@
...
+    path: "{{ image_dir }}"
+    file_type: file
+    recurse: yes
+  register: image_files
+
+- debug: var=image_files
+
+- name: Docker load images
+  command: docker load -i {{ item.path }}
+  with_items: "{{ image_files.files }}"
```
jeremyeder: i wonder if we need to somehow specify the forks value here. i could see a lot of failures if we fan out to N-nodes and the backend storage becomes saturated, leading to a domino effect of errors...docker is not good at handling extreme load.

image_provisioner/playbooks/roles/create-local-repo/vars/main.yaml
sjug: nginx is in EPEL I believe

image_provisioner/playbooks/roles/dns-client-deploy/tasks/main.yaml
sjug: We can move to the e2e solution, I just haven't integrated it yet

image_provisioner/playbooks/roles/dns-client-deploy/tasks/main.yaml
sjug: Will parameterize

image_provisioner/playbooks/roles/dns-server-deploy/vars/main.yaml
sjug: They're all vars?

image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
sjug: Nope, just kept the same value as before however we were previously just waiting the fixed timeout. Will double timeout.

image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
sjug: The SELinux task will happen either way, but it seems that `docker_storage_driver` must be set one way or another (overlay or devicemapper). There is no default.

image_provisioner/playbooks/roles/docker-spread-images/tasks/main.yaml
sjug: The this load task only puts stress on the local storage, do you think that is an issue? I haven't seen any errors in any of the deploys so far?

image_provisioner/playbooks/roles/dns-server-deploy/templates/dhcpd.conf.j2
sjug: Definitely still untested, can remove.

image_provisioner/playbooks/roles/docker-spread-images/tasks/main.yaml
jeremyeder: I don't think we have.  But remember we're using a pretty high-end rig.  Also "local" isn't always "local".  In our scenario, local is actually network.

image_provisioner/playbooks/bare_metal.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - create-local-repo
+    - docker-save
+    - dns-server-deploy
+
+- name: Allow root access
+  hosts: nodes
+  user: heat-admin
+  become: yes
+  roles:
+    - ssh-authorized-keys
```
sjug: Not sure if we should keep this? @jeremyeder

image_provisioner/inventory/hosts.BM
```diff
@@ -0,0 +1,48 @@
...
+
+[nodes]
+#10.2.4.10
+
+[nodes:vars]
+vm_install=False
+internal_image=False
+jump_node_ip=10.2.1.3
+# kickstart
+ssh_repo=/perf-dept/contents/ssh_keys/id_rsa_perf
```
jeremyeder: move sample inventory to private repo

image_provisioner/playbooks/bare_metal.yaml
jeremyeder: Why not?

jeremyeder: ?

sjug: Typically we build in the root password, ssh keys, etc into the cloud config. So maybe we wanted to do it a different way, in the deployment of the base machines?

jeremyeder: OK.  Just that all the private stuff needs to get out of this repo.

mffiedler:
@wabouhamad Please review

PR #193, opened 2017-04-12
Ansible playbook for setting up reliability cluster

vikaslaad:


reliability/setup.yml
```diff
@@ -0,0 +1,166 @@
+---
```
jeremyeder: wait -- we have implemented a register tools role in pbench.  please, please please use that instead so we can all rally around a single tool.  @chaitanyaenr can you please help here?

chaitanyaenr: @jeremyeder @vikaslaad  Here is a link to the playbook we have - https://github.com/distributed-system-analysis/pbench/tree/master/contrib/ansible/openshift. The labeling of the nodes starts from 0 instead of 1, I have a PR which fixes that - https://github.com/distributed-system-analysis/pbench/pull/551.

jeremyeder:
do we need this anymore?

reliability/setup.yml
```diff
@@ -0,0 +1,179 @@
...
+    - fetch:
+        src: /etc/origin/master/admin.kubeconfig
+        dest: runtime/keys
+        flat: yes
+    - fetch:
+        src: /etc/origin/master/ca.crt
+        dest: runtime/keys
+        flat: yes
+    - synchronize:
+        src: /usr/share/openshift/examples/quickstart-templates/*.*
```
wabouhamad: For containerized OCP installs the quickstart-templates are not available in this directory.  Instead we need to copy them from this directory on master:
/etc/origin/examples/quickstart-templates/*.*

PR #200, opened 2017-05-05
Added pbench-register-node-vertical-test-client.sh and nodeVertical-5…

wabouhamad:
…00-pods-ci.yaml for CI Automation


openshift_scalability/ci/contents/nodeVertical-500-pods-ci.yaml
```diff
@@ -0,0 +1,27 @@
+projects:
+  - num: 1
+    basename: clusterproject
+    ifexists: delete
```
mffiedler: Let's just update nodeVertical.yaml to add the ifexists: delete

mffiedler:
@wabouhamad I think we need to switch to an approach of using @chaitanyaenr pbench setup playbook and the ability in Flexy to launch a jump node.   What do you think?

openshift_scalability/ci/contents/nodeVertical-500-pods-ci.yaml
wabouhamad: @mffiedler I can update the nodeVertical.yaml file to add the ifexists: delete line.  We have been keeping copies of yaml config files in the ci/content subdir so scripts and config files are both contained in the ci dir.  Do we only want to keep scripts in the ci dir ?

wabouhamad:
@mffiedler I started to look at @chaitanyaenr's  pbench setup playbook after I wrote this script, which controls the timing between starting and stopping pbench collection, and does additional post processing to construct the pbench results URLs for all nodes.  I can cut and paste that log section into Polarion.  May be we can convert the pbench collection time management, running cluster-loader, and result post processing into a separate playbook.   Are you ok to have both approaches ?

jeremyeder:
Pinging to see if this conversation has occurred somewhere else offline?

wabouhamad:
had a conversation with @mffiedler.  The direction was to leverage @chaitanyaenr's pbench setup playbook during Flexy install.  This is done.   I updated my shell script to start pbench on the nodes where pbench was already registered, and proceed with Node Vertical test.  I'll be submitting an updated script in the next day or so

PR #201, opened 2017-05-07
Fix kernel upgrade repo and ansible 2.3.0.0 compatibility

sjug:
New kernel repo is actually a template not just a file to be copied over given the parameter.

The final repo-install step errors out if (vm_install) is not set in the inventory (which it's not by default). 

Finally in Ansible 2.3.0.0 using jinja templates in when statements has been deprecated so that has been resolved.


mffiedler:
Merging so that @wabouhamad can submit his PR

PR #203, opened 2017-05-11
support for sysbench running in containers

ekuric:
please check sysbench.adoc for full details
added below files : 

dockerfile centos6:  dockerfiles/centos6
dockerfile centos7:  dockerfiles/centos7
pbench helper script: runsys.sh
clusterloader parameters file: sysbench-parameters.yaml
sysbench template file : sysbench-template.json
readme : sysbench.adoc
script to run sysbench inside pods:  sysbenchose.sh


ekuric:
readme: https://github.com/ekuric/svt/blob/a6bef5f85e97668b65bb0fd9bb49b974fc64ff11/openshift_scalability/content/sysbench.adoc

mffiedler:
@ekuric  Sorry to take so long looking at this.   One comment:  The pattern we've started using (not sure if it is good, but it is in being used) is to create a directory under content for new tools and then put all of the content for that tool under that directory.   If it is not too much trouble, can you refactor so that all of the "content" files for this PR go under something like content/sysbench or content/runsys?    

Alternatively we could merge and then refactor if that is easier.

Examples of what I am talking about are content/centos-stress, content/logtest (see PR #214)

Let me know how you want to proceed.

ekuric:
@mffiedler I think that is good approach, #225 has that all

PR #205, opened 2017-05-15
Add option to run locally on master - no oc login needed

mffiedler:
New flag to bypass login.   Useful when running as system:admin on the master.


mffiedler:
@vikaslaad please take a look

PR #226, opened 2017-06-13
Use docker-storage-setup instead of manual setup

sjug:
Instead of setting up our docker storage manually, this uses docker-storage-setup. Needs a bit more testing with devicemapper before we can merge.

### DIFF
--- a/image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
@@ -1,44 +1,54 @@
 ---
-- name: set selinux to enforcing
-  selinux:
-    policy: targeted
-    state: enforcing
+#- name: set selinux to enforcing
+#  selinux:
+#    policy: targeted
+#    state: enforcing
 
-- name: copy docker-storage-setup config when storage driver is devicemapper
+- block:
+  - name: create docker config directory
+    file:
+      path: /root/.docker
+      state: directory
+  
+  - name: download docker login config.json
+    copy:
+      src: /root/{{ config_repo }}/image_provisioner/config.json
+      dest: /root/.docker/
+      remote_src: true
+  
+  - name: install docker
+    yum:
+      name: docker
+      state: latest
+  when: not atomic | default(False) | bool
+
+- name: stop docker service
+  systemd:
+    name: docker
+    state: stopped
+
+- name: reset docker storage
+  command: atomic storage reset
+  when: atomic | default(False) | bool
+
+- name: check that docker-storage has been deleted
+  stat:
+    path: /etc/sysconfig/docker-storage
+  register: ds_config
+
+- name: delete docker-storage if it still exists
+  file:
+    path: /etc/sysconfig/docker-storage
+    state: absent
+  when: ds_config.stat.exists == True
+
+- name: copy docker-storage-setup config
   template:
     src: etc/sysconfig/docker-storage-setup.j2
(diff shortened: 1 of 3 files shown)


image_provisioner/playbooks/roles/docker-config/templates/etc/sysconfig/docker-storage-setup.j2
```diff
@@ -1,2 +1,12 @@
-VG={{ docker_storage_vgname }}
-STORAGE_DRIVER={{ docker_storage_driver }}
+DEVS="{{ docker_storage_device }}"
+VG="{{ docker_storage_vgname }}"
+
+{% set storagedriver = docker_storage_driver | default('devicemapper') %}
```
sjug: Not really necessary to create new variable name, just adds default if unset

image_provisioner/playbooks/roles/docker-config/templates/etc/sysconfig/docker-storage-setup.j2
```diff
@@ -1,2 +1,12 @@
...
+STORAGE_DRIVER="{{ storagedriver }}"
+{% if storagedriver == 'devicemapper' %}
+DATA_SIZE=95%VG
```
sjug: Could also move to `100%FREE`

image_provisioner/playbooks/roles/docker-config/templates/etc/sysconfig/docker-storage-setup.j2
```diff
@@ -1,2 +1,12 @@
-VG={{ docker_storage_vgname }}
-STORAGE_DRIVER={{ docker_storage_driver }}
+DEVS="{{ docker_storage_device }}"
+VG="{{ docker_storage_vgname }}"
```
sjug: I don't think this needs to be exposed in the inventory

image_provisioner/playbooks/roles/docker-config/tasks/main.yaml
```diff
@@ -57,20 +45,16 @@
...
-     ignore_errors: true
-  
-   - name: exit if kernel version is outdated
-     fail: 
-       msg: "please update the kernel to 3.10.0-517 or later"
-     when: vercmp_out.rc == 12
+     failed_when: vercmp_out.rc == 12
   when: docker_storage_driver == "overlay2" or update_kernel == "true"
 
+# We can remove this but need to test without it
```
sjug: Still need to test without this block for devicemapper

sjug:
May need some changes for atomic host support, will test.

image_provisioner/playbooks/roles/docker-config/templates/etc/sysconfig/docker-storage-setup.j2
```diff
@@ -1,2 +1,15 @@
-VG={{ docker_storage_vgname }}
-STORAGE_DRIVER={{ docker_storage_driver }}
+{% if atomic == False %}
```
sjug: Need better condition with default

sjug:
Will finalize testing today, it seems that RHEL doesn't like the `CONTAINER_ROOT_*` params.

chaitanyaenr:
@sjug it's working on atomic for both devicemapper and overlay2.

sjug:
Thanks @chaitanyaenr, verified working on RHEL with both storage drivers (dm & overlay2) as well.

PR #227, opened 2017-06-13
Modified setup-image for atomic

sjug:
Not particularly a clean way to get it done from an efficiency perspective, the filenames prevent command reuse.
However this is functionally correct as the atomic qcow does come pre-partitioned with one device and two partitions. We only need to resize the second partition in this step.


sjug:
Also modified the setup-image so it can be run remotely on the same (destination machine) that will has libvirt on it. This saves the file transfer between `setup-image` and `provision_gold_images`.

PR #217, opened 2017-06-06
Add support to image provisioner to generate atomic image

chaitanyaenr:
This commit:
- generates atomic ami and qcow images when atomic variable is set
  to true.
- bakes in pbench-agent, controller and collectd images in to the
  image.
- Includes openshift templates to launch pbench-agent and collectd
  pods as daemonset
- Installs latest pbench using the rpm generated by Jenkins on a
  daily basis.
- Resizes volumes.

PR #228, opened 2017-06-13
Exclude ansible when performing os updates from ansible

chaitanyaenr:
Updating the ansible binary when a playbook is running will lead to
task failure.

### DIFF
--- a/image_provisioner/playbooks/roles/os-kickstart/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/os-kickstart/tasks/main.yaml
@@ -2,10 +2,12 @@
 - name: yum update
   yum: name=* state=latest
 
+- name: make sure ansible is installed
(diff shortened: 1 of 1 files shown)


sjug:
What conditions does that cause failure? I've not seen that.

chaitanyaenr:
this will cause failure when we don't have the latest ansible installed and try to update ansible from ansible playbook. Usually the os-kickstart installs packages on a remote machine that doesn't have ansible but when we already have ansible installed and try to update it, it fails. Take a look at https://github.com/ansible/ansible/issues/24059.

sjug:
Yeah but on what image does it update ansible?

chaitanyaenr:
on the host where we run all the roles

chaitanyaenr:
superseded by #217

PR #237, opened 2017-06-19
add python deps for openshift_scalability module

arcolife:
Add requirements.txt file for python scripts used in svt, and a naive helper script to install 'em.

### DIFF
--- a/openshift_scalability/prerequisites.sh
+++ b/openshift_scalability/prerequisites.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+yum install gcc libyaml python-virtualenv
+
+virtualenv venv
+source venv/bin/activate
+
+pip install -r requirements.txt
+
+# run test
+python utils.py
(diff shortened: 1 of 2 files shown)


jeremyeder:
Hmm, is this really all required by svt at this point? Are all of these requirements baked into our gold images?

arcolife:
I see that svt repo is cloned in [this playbook](https://github.com/openshift/svt/blob/master/image_provisioner/playbooks/provision_gold_images.yaml#L12-L16) but I don't get the connection between gold images and wanting to individually run cluster-loader scripts?

So every time someone clones svt and runs cluster-loader, they'd need to have these requirements satisfied in order to use https://github.com/openshift/svt/blob/master/openshift_scalability/utils.py which is imported by https://github.com/openshift/svt/blob/master/openshift_scalability/cluster-loader.py

I've recently seen another person facing the same issue and decided to send a PR for it. Am I seeing this right or is there another way to use it?

mffiedler:
Closing in favor of ongoing work in the golang cluster-loader.

PR #243, opened 2017-06-26
Parameterize instance_tag to avoid conflicts

chaitanyaenr:
- Setting the instance_tag in the inventory before building an AMI will
  avoid conflicts and enable us to build multiple images in parallel.

- A sample use case would be using Jenkins to build to build RHEL and 
  atomic images at once.

### DIFF
--- a/image_provisioner/inventory/hosts.AWS
+++ b/image_provisioner/inventory/hosts.AWS
@@ -11,6 +11,7 @@ group_id=
 instance_type=m4.xlarge
 vpc_subnet_id=
 base_image=
+instance_tag=
 # repo install
 rhel_int_repo_hostname=
 # os kickstart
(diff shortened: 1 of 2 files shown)


chaitanyaenr:
@vikaslaad PTAL.

image_provisioner/playbooks/build_ami.yaml
```diff
@@ -25,9 +25,9 @@
       user_data: "{{ lookup('file', '../cloud-init/user-data-aws') }}"
       exact_count: 1
       count_tag:
-         Name: ami_builder
+         Name: "{{ instance_tag | default('ami_builder') }}"
```
sjug: This only works when `instance_tag` is not defined.

image_provisioner/inventory/hosts.AWS
```diff
@@ -11,6 +11,7 @@ group_id=
 instance_type=m4.xlarge
 vpc_subnet_id=
 base_image=
+instance_tag=
```
sjug: With this set in the inventory the instance name will be <empty>

PR #249, opened 2017-07-10
Parameterize ami_name to avoid conflicts

chaitanyaenr:
This is needed for Jenkins Jobs since they get triggered once a new
puddle is found and builds atomic and rhel ami's in parallel. We need
to have different ami names to differentiate between rhel and atomic.


chaitanyaenr:
@vikaslaad  @jeremyeder PTAL.

sjug:
@chaitanyaenr Jeremy most likely does not have the cycles to code review all your PRs. Please tag @openshift/svt moving forward.

PR #260, opened 2017-07-18
Syntax error due to indentation

jhadvig:
While playing around reliability tests, I've hit this indentation error. Not sure if its due my version of ansible (ansible-2.3.1.0-1.el7) but after this I was able to execute the `setup.yml` playbook.
@mffiedler PTAL


mffiedler:
@vikaslaad Please take a look

jhadvig:
Adding addtional fixes for the setup.yml

PR #264, opened 2017-07-23
configurable yaml test file

mrsiano:
@vikaschoudhary16 
we should be able to manage the yaml test file form outside, instead of editing hard coded lines.

### DIFF
--- a/networking/synthetic/stac-prepare-nodes.py
+++ b/networking/synthetic/stac-prepare-nodes.py
@@ -82,20 +82,22 @@ def parse_args():
                         dest='iface_name',
                         help='interace name on nodes through which stac test traffic will run')
 
-    parser.add_argument('-t',
-                        '--test-file',
-                        required=True,
-                        dest='test_file',
-                        help='a yaml test file')
+    parser.add_argument('-p',
+                        '--tracing',
+                        required=False,
+                        dest='pbd_tracing',
+                        help='profile the app with pdb tracing')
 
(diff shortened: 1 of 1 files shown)


mrsiano:
@jeremyeder

networking/synthetic/stac-prepare-nodes.py
```diff
@@ -82,14 +82,28 @@ def parse_args():
...
+                        required=False,
+                        dest='pbd_tracing',
+                        help='profile the app with pdb tracing')
+
     return parser.parse_args()
 
 
 def main():
     args = parse_args()
-    nettest = NetworkTest('stac-test.yaml')
```
vikaschoudhary16: @mrsiano IMO, only file name correction, `stac-prepare-nodes.yaml` instead of `stac-test.yaml`, should be enough.
I think we should not make it configurable to keep knobs to a minimal for ease of use. I see this as an internal file which user should not be aware about. Any test case related configuration is maintained at github repo, which is passed as a command line variable. And because that repo path will change from user to user, so it makes sense to pass that as a variable. This file name will not change like that.

mrsiano: done,
please review

vikaschoudhary16:
LGTM.

mrsiano:
@vikaschoudhary16 @jeremyeder 
Please merge

mrsiano:
@openshift/svt

PR #265, opened 2017-07-24
Changes for tolerating bad pods

ravisantoshgudimetla:
- This is to ensure that cluster-loader won't wait till all the pods come to running state if "-t" option is provided. 
- By default if -t is not provided cluster-loader will run as usual and will wait. 
- Three examples for bad pods which Derek gave are included in content folder.

cc @jeremyeder @sjug

### DIFF
--- a/openshift_scalability/README.md
+++ b/openshift_scalability/README.md
@@ -104,6 +104,7 @@ tuningsets:
 > * The ***Tuning parameters*** have following function:
 >  * ***stepping*** : This feature makes sure that after each "stepsize" pod/template requests are submitted, they enter the "Running" state. After all the pods/templates in the given step are Running, then there is a delay = "pause" , before the next step.
 >  * ***rate_limit*** : This makes sure that there is a delay of "rate_limit.delay" between each pod/template request submission.
+>  * ***--tolerate-bad-pods*** : This option ensures that cluster-loader utility won't wait till all the pods would come to running state. This is useful when we intentionally give bad pods which are bound to fail. By default the utility waits till all the pods come to running state.
 
 ```
 This Config file will create the following objects :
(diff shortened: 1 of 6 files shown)


openshift_scalability/README.md
```diff
@@ -104,6 +104,7 @@ tuningsets:
 > * The ***Tuning parameters*** have following function:
 >  * ***stepping*** : This feature makes sure that after each "stepsize" pod/template requests are submitted, they enter the "Running" state. After all the pods/templates in the given step are Running, then there is a delay = "pause" , before the next step.
 >  * ***rate_limit*** : This makes sure that there is a delay of "rate_limit.delay" between each pod/template request submission.
+>  * ***--tolerate-bad-pods*** : This option ensures that cluster-loader utility won't wait till all the pods would come to running state. This is useful when we intentionally give bad pods which are bound to fail. By default the utility waits till all the pods come to running state.
```
sjug: This is not a tuning parameter, please put it in the correct section of the README.

openshift_scalability/utils.py
```diff
@@ -825,14 +824,15 @@ def pod_handler(inputpods, globalvars):
...
         create_pods(pod_config, num,storagetype, globalvars)
-
-    if len(globalvars["pend_pods"]) > 0:
-        pod_data(globalvars)
-
-    if "podtuningset" in globalvars:
-        del(globalvars["podtuningset"])
-        del(globalvars["totalpods"])
-    del(globalvars["pend_pods"])
+    tolerate_bad_pods = globalvars["tolerate"]
```
sjug: Local variables should be set at the beginning of the function, with the rest of them.

ravisantoshgudimetla:
@sjug PTAL.

ravisantoshgudimetla:
Also, this flag needs to go into golang version of clusterloader.

sjug:
/lgtm

We are moving away from flags and using the config instead in golang clusterloader, but yes it will also need this functionality.

ravisantoshgudimetla:
@sjug /lgtm won't work. So, we need your manual merge.

ravisantoshgudimetla:
Thanks @mffiedler for merging.

PR #287, opened 2017-08-13
pbench log parser tool

sjug:
Not exactly sure where this should go.

This tool calculates statistics based on the raw pbench data, and outputs a summary csv.

During execution the stdout logging is still very verbose.

### DIFF
--- a/pbench_logparse/README.md
+++ b/pbench_logparse/README.md
@@ -1,0 +1,32 @@
+# pbench_logparse
+
+This golang tool was created to parse the output of time series data created by [pbench](https://github.com/distributed-system-analysis/pbench) in CSV format.
+
+```
+Usage of ./pbench_logparse:
+  -blkdev value
+        List of block devices
+  -i string
+        pbench run result directory to parse (default "/var/lib/pbench-agent/benchmark_result/tools-default/")
+  -netdev value
+        List of network devices
+  -o string
+        output directory for parsed CSV result data (default "/tmp/")
+  -proc string
+        list of processes to gather (default "openshift_start_master_api,openshift_start_master_controll,openshift_start_node,/etcd")
+```
+
+Example command:
+```
+./pbench_logparse ~/work/pbench-result/tools-default/ -o ~/data/ -blkdev vda-write -blkdev xvdb -netdev eth0-rx -netdev eth0-tx
+```
+
+`blkdev` represents a single block device name, to add more than one block device, you will need to pass the flag again per device, as above
+
+`i` is the input directory, it must point to the parent of the host data, which is `.../tools-default/`
+
+`o` is the output directory, it can be any directory that includes a trailing slash (dirname/)
+
+`netdev` represents a single network device name, to add more than more network device, you will need to pass the flag again per device, as above
+
+`proc` is a comma-separated list of process names to extract results for, avoid spaces
--- a/pbench_logparse/main.go
+++ b/pbench_logparse/main.go
@@ -1,0 +1,321 @@
+package main
+
+import (
+	"bufio"
+	"encoding/csv"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/sjug/go-logparse/stats"
+)
+
+var searchDir, resultDir, processes string
+var processList []string
+var blockDevices, netDevices stringSlice
+var fileHeader = map[string][]string{}
+
+type stringSlice []string
+
+type host struct {
+	kind      string
+	resultDir string
+	results   []resultType
+}
+
+type resultType struct {
+	kind, path           string
+	min, max, avg, pct95 float64
+}
+
+func (s *stringSlice) String() string {
+	return fmt.Sprintf("%v", *s)
+}
+
+func (s *stringSlice) Set(value string) error {
+	*s = append(*s, value)
+	return nil
+}
+
+func initFlags() {
+	flag.StringVar(&searchDir, "i", "/var/lib/pbench-agent/benchmark_result/tools-default/", "pbench run result directory to parse")
+	flag.StringVar(&resultDir, "o", "/tmp/", "output directory for parsed CSV result data")
+	flag.StringVar(&processes, "proc", "openshift_start_master_api,openshift_start_master_controll,openshift_start_node,/etcd", "list of processes to gather")
+	flag.Var(&blockDevices, "blkdev", "List of block devices")
+	flag.Var(&netDevices, "netdev", "List of network devices")
+	flag.Parse()
+}
+
+func main() {
+	var hosts []host
+	// This regexp matches the prefix to each pbench host result directory name
+	// which indicates host type. (ie. svt-master-1:pbench-benchmark-001/)
+	hostRegex := regexp.MustCompile(`svt[_-][elmn]\w*[_-]\d`)
+	initFlags()
+
+	// Check if no flags were passed, print help
+	if flag.NFlag() == 0 {
+		flag.PrintDefaults()
+		return
+	}
+
+	// Ensure output directory has a trailing slash
+	if string(resultDir[len(resultDir)-1]) != "/" {
+		resultDir = resultDir + "/"
+	}
+
+	// If no block devices were passed, don't add to search
+	if len(blockDevices) > 0 {
+		fileHeader["disk_IOPS.csv"] = []string(blockDevices)
+	}
+
+	// If no network devices were passed, don't add to search
+	if len(netDevices) > 0 {
+		fileHeader["network_l2_network_packets_sec.csv"] = []string(netDevices)
+		fileHeader["network_l2_network_Mbits_sec.csv"] = []string(netDevices)
+	}
+
+	processList = strings.Split(processes, ",")
+	// If no process names were passed, don't add to search
+	if len(processList) > 0 {
+		fileHeader["cpu_usage_percent_cpu.csv"] = processList
+		fileHeader["memory_usage_resident_set_size.csv"] = processList
+	}
+
+	// Return director listing of searchDir
+	dirList, err := ioutil.ReadDir(searchDir)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Iterate over directory contents
+	for _, item := range dirList {
+		// Match subdirectory that follows our pattern
+		if hostRegex.MatchString(item.Name()) && item.IsDir() {
+			kind := strings.Split(item.Name(), ":")
+			newHost := host{
+				kind:      kind[0],
+				resultDir: searchDir + item.Name(),
+			}
+			hosts = append(hosts, newHost)
+		}
+	}
+
+	// Maps are not ordered, create ordered slice of keys and sort
+	// This ensures that the file output is identical between execution
+	keys := []string{}
+	for k := range fileHeader {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	// Iterate over all known hosts
+	for i, host := range hosts {
+		fmt.Printf("Host: %+v\n", host)
+		// Find each raw data CSV
+		for _, key := range keys {
+			fmt.Println(key)
+			fileList := findFile(host.resultDir, key)
+			// findFile returns slice, though there should only be one file
+			for _, file := range fileList {
+				// Parse file into 2d-string slice
+				result, err := readCSV(file)
+				if err != nil {
+					fmt.Printf("Error reading %v: %v\n", file, err)
+					continue
+				}
+				// In a single file we have multiple headers to extract
+				for _, header := range fileHeader[key] {
+					// Extract single column of data that we want
+					newResult, err := newSlice(result, header)
+					if err != nil {
+						//need to keep list of columns same for all types
+						//continue
+						fmt.Printf("newSlice returned error: %v\n", err)
+					}
+
+					// Mutate host to add calcuated stats to object
+					hosts[i].addResult(newResult, file, header)
+
+				}
+				fmt.Printf("CALLER Host populated: %+v\n", hosts[i])
+
+			}
+		}
+	}
+
+	err = writeCSV(keys, hosts)
+	if err != nil {
+		fmt.Printf("Error writing CSV: %v", err)
+	}
+}
+
+func writeCSV(keys []string, hosts []host) error {
+	csvFile, err := os.Create(resultDir + "out.csv")
+	if err != nil {
+		return err
+	}
+	defer csvFile.Close()
+
+	// Write test CSV data to stdout
+	writer := csv.NewWriter(csvFile)
+	defer writer.Flush()
+
+	// Create header & write
+	header := createHeaders(keys)
+	for _, h := range header {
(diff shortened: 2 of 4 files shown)


pbench_logparse/README.md
```diff
@@ -0,0 +1,13 @@
+# pbench_logparse
+
+This golang tool was created to parse the output of time series data in CSV.
+
+```
+Usage of ./go-logparse:
+  -i string
+        pbench run result directory to parse (default "/home/jugs/work/20170330_2Knodes_5Kprojects_nvme_etcd_a/tools-default/")
```
jeremyeder: wat

pbench_logparse/main.go
```diff
@@ -0,0 +1,279 @@
...
+	results   []resultType
+}
+
+type resultType struct {
+	kind, path           string
+	min, max, avg, pct95 float64
+}
+
+var fileHeader = map[string][]string{
+	"disk_IOPS.csv":                      {"nvme0n1-write", "vda-write"},
```
jeremyeder: How to make this extensible without re-compiling?

pbench_logparse/main.go
```diff
@@ -0,0 +1,279 @@
...
+	"cpu_usage_percent_cpu.csv":          {"openshift_start_master_api", "openshift_start_master_controll", "openshift_start_node", "\\/etcd"},
+	"memory_usage_resident_set_size.csv": {"openshift_start_master_api", "openshift_start_master_controll", "openshift_start_node", "\\/etcd"},
+	"network_l2_network_packets_sec.csv": {"eth0-rx", "eth0-tx"},
```
jeremyeder: device name hardcoded :-(

pbench_logparse/main.go
```diff
@@ -0,0 +1,279 @@
...
+	"network_l2_network_Mbits_sec.csv":   {"eth0-rx", "eth0-tx"},
+}
+
+func initFlags() {
+	flag.StringVar(&searchDir, "i", "/home/jugs/work/20170330_2Knodes_5Kprojects_nvme_etcd_a/tools-default/", "pbench run result directory to parse")
```
jeremyeder: Might want to clean this up.

pbench_logparse/main.go
```diff
@@ -0,0 +1,279 @@
...
+	flag.StringVar(&resultDir, "o", "/tmp/", "directory to output parsed CSV result data")
+	flag.Parse()
+}
+
+func main() {
+	var hosts []host
+	hostRegex := regexp.MustCompile(`svt[_-][elmn]\w*[_-]\d`)
```
jeremyeder: Or make this regex somehow configurable?

jeremyeder:
@ravisantoshgudimetla @mffiedler

pbench_logparse/README.md
```diff
@@ -0,0 +1,32 @@
+# pbench_logparse
+
+This golang tool was created to parse the output of time series data in CSV.
```
ravisantoshgudimetla: As of now, can parse from pbench output alone. Please update accordingly.

pbench_logparse/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+```
+Usage of ./pbench_logparse:
+  -blkdev value
+        List of block devices
+  -i string
+        pbench run result directory to parse (default "/var/lib/pbench-results/test1/tools-default/")
```
ravisantoshgudimetla: I guess, there is no need to have a default value here.

pbench_logparse/README.md
```diff
@@ -0,0 +1,32 @@
...
+  -netdev value
+        List of network devices
+  -o string
+        directory to output parsed CSV result data (default "/tmp/")
```
ravisantoshgudimetla: Directory which stores the output file that has results.

pbench_logparse/main.go
```diff
@@ -0,0 +1,317 @@
...
+	return fmt.Sprintf("%v", *s)
+}
+
+func (s *stringSlice) Set(value string) error {
+	*s = append(*s, value)
+	return nil
+}
+
+func initFlags() {
+	flag.StringVar(&searchDir, "i", "/var/lib/pbench-results/test1/tools-default/", "pbench run result directory to parse")
```
ravisantoshgudimetla: nit: Please remove the default value here.

pbench_logparse/main.go
```diff
@@ -0,0 +1,317 @@
...
+	flag.StringVar(&resultDir, "o", "/tmp/", "directory to output parsed CSV result data")
+	flag.StringVar(&processes, "proc", "openshift_start_master_api,openshift_start_master_controll,openshift_start_node,/etcd", "list of processes to gather")
+	flag.Var(&blockDevices, "blkdev", "List of block devices")
```
ravisantoshgudimetla: Any specific reason for not having a comma separated values here? In fact, if we combine all the input params into a struct, it could be passed along easily. WDYT?

pbench_logparse/main.go
ravisantoshgudimetla: +1

pbench_logparse/README.md
sjug: The idea is to provide an example directory, having it doesn't hurt anything but it exists for documentation.

pbench_logparse/main.go
sjug: I don't see any benefit in removal

pbench_logparse/README.md
sjug: I've reworded it for better clarity

pbench_logparse/main.go
sjug: The line you've commented on does not have comma separated values?
Combine them how? Passed along to what?

pbench_logparse/README.md
sjug: Sure

pbench_logparse/main.go
```diff
@@ -0,0 +1,317 @@
...
+
+	"github.com/sjug/go-logparse/stats"
+)
+
+var searchDir, resultDir, processes string
+var processList []string
+var blockDevices, netDevices stringSlice
+var fileHeader = map[string][]string{}
+
+type stringSlice []string
```
ravisantoshgudimetla: Any specific reason for having a custom type here?

pbench_logparse/main.go
```diff
@@ -0,0 +1,317 @@
...
+	resultDir string
+	results   []resultType
+}
+
+type resultType struct {
+	kind, path           string
+	min, max, avg, pct95 float64
+}
+
+func (s *stringSlice) String() string {
```
ravisantoshgudimetla: Why do we need this fn?

pbench_logparse/main.go
```diff
@@ -0,0 +1,317 @@
...
+	return fmt.Sprintf("%v", *s)
+}
+
+func (s *stringSlice) Set(value string) error {
```
ravisantoshgudimetla: Same as above comment

pbench_logparse/main.go
sjug: I like how you're explaining the benefit of what I've done on the line directly above, thanks. The reason I didn't do the same thing for both is because usually we don't have many block devices, nor network devices, where we do have a large number of process names. Also, it was a fun experiment to extend the flag interface.

pbench_logparse/main.go
sjug: Because I've created a custom flag type for netdev and blkdev?

pbench_logparse/main.go
sjug: Look up the flag interface.

pbench_logparse/main.go
sjug: Same comment as above

pbench_logparse/main.go
ravisantoshgudimetla: > The reason I didn't do the same thing for both is because usually we don't have many block devices,

Ok but probably from user-experience perspective, it would cause inconsistency.

> Also, it was a fun experiment to extend the flag interface.
:+1:

pbench_logparse/main.go
ravisantoshgudimetla: If not, please add a comment what the regex does and why we need it.

pbench_logparse/main.go
sjug: I agree with you 100% on the inconsistency, I'll discuss this point tomorrow.

pbench_logparse/main.go
ravisantoshgudimetla: Ok, so my question is why do we want to create a custom flag in first place? If we have comma separated value, we could easily pass them around as slice.

pbench_logparse/main.go
sjug: Added

pbench_logparse/main.go
sjug: How do we pass a comma separated value as a slice?

pbench_logparse/main.go
```diff
@@ -0,0 +1,319 @@
...
+			hosts = append(hosts, newHost)
+		}
+	}
+
+	// Maps are not ordered, create ordered slice of keys
+	keys := []string{}
+	for k := range fileHeader {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
ravisantoshgudimetla: Any specific reason for sorting here?

pbench_logparse/main.go
```diff
@@ -0,0 +1,319 @@
...
+
+	// Iterate over all known hosts
+	for i, host := range hosts {
+		fmt.Printf("Host: %+v\n", host)
```
ravisantoshgudimetla: I guess, we don't need this, if needed, we can have a logger.

pbench_logparse/main.go
sjug: So they're not in random order every time?

pbench_logparse/main.go
sjug: Yes, like I said in my PR message:
> During execution the stdout logging is still very verbose.

I can get rid of all the verbosity, not sure we want to quite yet.

pbench_logparse/main.go
ravisantoshgudimetla: But do we care?

pbench_logparse/main.go
```diff
@@ -0,0 +1,319 @@
...
+		// Find each raw data CSV
+		for _, key := range keys {
+			fmt.Println(key)
+			fileList := findFile(host.resultDir, key)
```
ravisantoshgudimetla: Why slice, if there is more than one value, shouldn't we log an error?

pbench_logparse/main.go
```diff
@@ -0,0 +1,319 @@
...
+	writer := csv.NewWriter(csvFile)
+	defer writer.Flush()
+
+	// Create header & write
+	header := createHeaders(keys)
+	for _, h := range header {
+		writer.Write(h)
+	}
+
+	// TODO: Maybe use reflection to get these fields instead
```
ravisantoshgudimetla: Or let the user decide what he/she wants, probably adding a flag won't hurt.

pbench_logparse/main.go
```diff
@@ -0,0 +1,319 @@
...
+			writer.Write(hosts[i].toSlice(v))
+		}
+	}
+	return nil
+}
+
+func createHeaders(keys []string) (header [][]string) {
+	empty := []string{""}
+	header = append(header, empty)
+	header = append(header, empty)
```
ravisantoshgudimetla: Is this a duplicate line? Also, just a nit, there is no need to declare empty variable.

pbench_logparse/main.go
```diff
@@ -0,0 +1,319 @@
...
+	for _, key := range keys {
+		k := strings.Split(key, ".")
```
ravisantoshgudimetla: Please add a comment why we are doing this split at `.`

pbench_logparse/main.go
sjug: Doesn't really matter? I guess the function is called `findFile` singular. Could just throw an error if the `len(fileList) > 1` if we really wanted?

pbench_logparse/main.go
ravisantoshgudimetla: I meant the way you did it earlier. Have a comma seperated string and then split it based on comma to create a slice.

pbench_logparse/main.go
ravisantoshgudimetla: Just for the sake of code tightening

pbench_logparse/main.go
sjug: This is for spacing of headers.
`empty := []string{""}` creates an allocated slice with 1 element, which I use for spacing
`empty := []string{}` creates an allocated slice with 0 elements, not useful

I need two empty CSV fields so I append the empty fields twice. I suppose I can just declare it explicitly instead.

pbench_logparse/main.go
sjug: In what scenario would someone not want all of these statistics?

pbench_logparse/main.go
sjug: Sure

pbench_logparse/main.go
ravisantoshgudimetla: Yes, you can add 2 empty elements in slice if you need.

pbench_logparse/main.go
sjug: Yes

pbench_logparse/stats/stats_test.go
```diff
@@ -0,0 +1,62 @@
...
+	values []float64
+	sum    float64
+	mean   float64
+	min    float64
+	max    float64
+	p95    float64
+}
+
+var tests = []teststruct{
+	// Floating precision error result kept intact for percentile result, rounding done in print, not in Percentile function
```
ravisantoshgudimetla: If possible add some -ve tests as well.

pbench_logparse/main.go
sjug: Sure

pbench_logparse/main.go
sjug: Indeed

pbench_logparse/main.go
ravisantoshgudimetla: one use-case was last week, when I wanted only the median because I wanted to have comparison point with data collected from online.

pbench_logparse/main.go
ravisantoshgudimetla: Cool, thanks for information. Please add a comment if possible.

pbench_logparse/main.go
sjug: I think you mean the mean? The tool doesn't calculate median?
So the goal of specifying type of statistic would be to cut the run time from 3s to 1s?

ravisantoshgudimetla: Yup, sorry about the "median", "mean" confusion. It does 2 things
- Reduces computation time as you told earlier.
- Increases flexibility for end-user.

sjug: Will bring this up as well tomorrow for the final word.

pbench_logparse/main.go
sjug: Sure.

mffiedler:
Based on the latest comment, merging and further improvements can be done in follow-up work

PR #297, opened 2017-09-06
Added dependencies to svt readme

bfallonf:
As per conversation in this BZ:

https://bugzilla.redhat.com/show_bug.cgi?id=1456234

cc @jeremyeder

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,9 @@ Unsurprisingly, performance analysis and tuning in the container and container-o
 
 Further tests quantify application performance when running in a container hosted by OpenShift, as well as measure reliability over time, searching for things like memory leaks.
 
+## IMPORTANT 
+While the tests in this repository are used by the Red Hat team to measure performance, these are **not supported in any OpenShift environments**, and Red Hat support services cannot provide assistance with any problems.
+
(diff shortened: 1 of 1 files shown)


jeremyeder:
For context, a customer requested more detail on building images to transfer into a disconnected environment.

@bfallonf the only thing I'd say is to make the language strong around how this is not supported in any way shape or form, regardless of whether it is production environment or not (In other words, you cannot call CEE and open a ticket against it).

That said, we'd like feedback via Issues and especially PRs to this repo.

bfallonf:
Thanks @jeremyeder . I've updated the paragraph. Let me know if you have any suggestions to make it stronger, etc. I also changed the bit at the bottom asking for feedback to be more precise.

Also, let me know if this needs labels or whatever, or if it needs more.

jeremyeder:
LGTM, long term goal is to migrate our tests into the Origin repo (which has already begun).  Those will be supported.

bfallonf:
@jeremyeder Oops. Too used to the openshift-docs repo markup... I made it a heading, but smaller. Let me know if there's anything else. Thanks. Merge away if you like?

PR #298, opened 2017-09-07
Update networking tests to collect pbench data for second remote.

schituku:


mffiedler:
@schituku Is this a change we still want to make?   Seems like a non-ordinary case?   Or that it should be done generically with some sort of an array vs. adding _two.   Let me know what you think.

schituku:
This change was done to collect perf data for the IpSec evaluation. As a stop gap solution to expedite ipSec testing I added this. IpSec will not be a regular test it was more of one time evaluation. so, Yes this is a non-oridinary case and we can discard this PR now.

mffiedler:
Closing this PR per @schituku comment.

PR #299, opened 2017-09-11
Add cri-o support to image provisioner

chaitanyaenr:
This commit:
- Adds support to use cri-o as runtime
- Mounts the volume at /var/lib and adds more storage needed to
  store both docker and cri-o images.
- Copies openshift images from docker storage to CRI-O storage.

### DIFF
--- a/image_provisioner/playbooks/build_ami.yaml
+++ b/image_provisioner/playbooks/build_ami.yaml
@@ -24,7 +24,7 @@
       volumes:
         - device_name: "{{ '/dev/sda1' if atomic | default(False) else '/dev/sdb' }}"
           volume_type: gp2
-          volume_size: 50
+          volume_size: "{{ volume_size | default(70) }}"
           delete_on_termination: true
       wait: yes
       user_data: "{{ lookup('file', '../cloud-init/user-data-aws') }}"
@@ -59,6 +59,7 @@
     - { role: os-kickstart }
     - { role: clone-repos, when: not atomic | default(False) | bool }
     - { role: docker-config }
+    - { role: cri-o, when: crio | default(False) | bool }
     - { role: pbench-config, when: not atomic | default(False) | bool }
     - { role: aos-ansible }
     - { role: ansible-update, when: not atomic | default(False) | bool }
--- a/image_provisioner/playbooks/roles/aos-ansible/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/aos-ansible/tasks/main.yaml
@@ -37,3 +37,20 @@
     name: iptables 
     state: started 
     enabled: yes
+
+- block:
+   - name: check if skopeo is installed
+     command: which skopeo
+     register: skopeo_status
+
+   - name: Install skopeo if not installed
+     yum:
+       name: skopeo
+       state: latest
+     when: not atomic | default(False) | bool and skopeo_status.rc != 0
+
+   - name: copy images to container storage
+     shell: for image in $(docker images {{ image_repository }}  --format {% raw %} '{{.Repository}}:{{.Tag}}' {% endraw %}); do \
+                skopeo copy docker-daemon:$image containers-storage:$image; done
+     when: skopeo_status.rc == 0
+  when: crio | default(False) | bool and crio_copy_images | default(False) | bool
--- a/image_provisioner/playbooks/roles/cri-o/files/crio-bridge.conf
+++ b/image_provisioner/playbooks/roles/cri-o/files/crio-bridge.conf
@@ -1,0 +1,15 @@
+{
+    "cniVersion": "0.2.0",
+    "name": "crio-bridge",
+    "type": "bridge",
+    "bridge": "cni0",
+    "isGateway": true,
+    "ipMasq": true,
+    "ipam": {
+        "type": "host-local",
+        "subnet": "10.88.0.0/16",
+        "routes": [
+            { "dst": "0.0.0.0/0" }
+        ]
+    }
+}
--- a/image_provisioner/playbooks/roles/cri-o/tasks/main.yml
+++ b/image_provisioner/playbooks/roles/cri-o/tasks/main.yml
@@ -1,0 +1,91 @@
+---
+- name: Ensure container-selinux, atomic and runc are installed
+  package:
+     name: "{{ item }}"
+     state: present
+  with_items:
+     - container-selinux
+     - atomic
+     - runc
+  when: not atomic | default(False) | bool
+
+- name: Use RHEL based image
+  set_fact:
+    crio_image_prepend: "{{ crio_registry }}"
+    crio_image_name: "cri-o"
+
+- name: Set the full image name
+  set_fact:
+    crio_image: "{{ crio_image_prepend }}/{{ crio_image_name }}:latest"
+
+- name: Pre-pull CRI-O System Container image
+  atomic_image:
+     name: "{{ crio_image }}"
+     state: latest
+     backend: ostree
+
+- name: Install CRI-O container
+  command: "atomic install --system-package no --system {{ crio_image }}"
+
+- name: Ensure that selinux is set to true in the config
+  lineinfile:
+    path: /etc/crio/crio.conf
+    regexp: '^selinux'
+    line: 'selinux = true'
+
+- name: Ensure that systemd is the cgroup manager
+  lineinfile:
+    path: /etc/crio/crio.conf
+    regexp: '^cgroup_manager'
+    line: "cgroup_manager = \"systemd\""
+
+- name: Ensure that file locking is set to false
+  lineinfile:
+    path: /etc/crio/crio.conf
+    regexp: '^file_locking'
+    line: 'file_locking = false'
+
+- name: override overlay2 kernel check
+  lineinfile:
+    path: /etc/crio/crio.conf
+    insertafter: '^storage_option'
+    line: "\t\"overlay2.override_kernel_check=1\""
+    state: present
+
+- name: create cni directory to store network settings
+  file:
+    path: "{{ item }}"
+    state: directory
+  with_items:
+     - /etc/cni/net.d
+     - /opt/cni/net.d
+
+- name: copy crio-bridge conf
+  copy:
+    src: crio-bridge.conf
+    dest: /etc/cni/net.d/10-mynet.conf
+
+- name: copy loopback conf
+  copy:
+    src: loopback.conf
+    dest: /etc/cni/net.d/99-loopback.conf
+
+- name: override overlay2 kernel check
+  lineinfile:
+    path: /etc/containers/storage.conf
+    insertafter: '^\[storage.options\]'
+    line: "override_kernel_check=\"true\""
+
+- name: Start the CRI-O service
+  systemd:
+    name: "cri-o"
+    enabled: yes
+    state: started
+    daemon_reload: yes
+
+- name: systemctl status
+  command: systemctl status cri-o
+  register: crio_status
+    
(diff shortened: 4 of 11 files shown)


image_provisioner/playbooks/roles/aos-ansible/tasks/main.yaml
```diff
@@ -37,3 +37,18 @@
...
+- block:
+   - name: check if skopeo is installed
+     command: rpm -q skopeo
+     register: skopeo_status
+
+   - type: "script"
+     title: "copy openshift images from docker storage to CRI-O storage"
+     timeout: 10800
+     script: |-
+          for i in $(docker images registry.ops.openshift.com/openshift3/* --format "{{.ID}}: {{.Repository}}"); do
```
jeremyeder: this will break soon.  the registry url is changing.  any chance this can somehow be parameterized?

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Ensure runc is installed
+      package:
+        name: runc
+        state: present
+  when: not atomic | default(False) | bool
+
+# change the registery to registry.access.redhat.com once there is an official image is available
+- name: Use RHEL based image
+  set_fact:
+    crio_image_prepend: "docker.io/gscrivano"
```
jeremyeder: this has to be parameterized as well...

image_provisioner/playbooks/roles/cri-o/templates/crio.conf.j2
```diff
@@ -0,0 +1,123 @@
+# The "crio" table contains all of the server options.
```
jeremyeder: Having the whole config file in our git is going to get very ugly.  Suggest thinking this through more...there will be no way to keep up with the upstream config file options.  Perhaps only include things that we modify and use lineinfile ansible module to juggle it?

image_provisioner/playbooks/roles/docker-config/templates/etc/sysconfig/docker-storage-setup.j2
```diff
@@ -9,7 +9,7 @@ DATA_SIZE=100%FREE
 {% elif docker_storage_driver == 'overlay2' %}
 {% if atomic | bool %}
 CONTAINER_ROOT_LV_NAME=docker-root-lv
-CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/docker
+CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib
```
jeremyeder: What takes care of mounting an overlay volume at /var/lib?

image_provisioner/playbooks/roles/aos-ansible/tasks/main.yaml
chaitanyaenr: Done.

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
chaitanyaenr: Done.

image_provisioner/playbooks/roles/docker-config/templates/etc/sysconfig/docker-storage-setup.j2
chaitanyaenr: docker when started starts the docker-storage-setup which takes care of it.

sjug: Why are we making all of `/var/lib` into an overlay2 lv? Docker storage setup should be for docker only AFAIK, CRI-O does its own thing to set up `/var/lib/containers`.

chaitanyaenr: docker storage setup will assign all of the free space to /var/lib/docker, so /var/lib/containers will not have much storage. Making all of /var/lib in to an overlay2 lv will solve the issue.

mffiedler:
@jeremyeder @sjug Concerns addressed?

jeremyeder:
si`

image_provisioner/playbooks/build_ami.yaml
```diff
@@ -24,7 +24,7 @@
       volumes:
         - device_name: "{{ '/dev/sda1' if atomic | default(False) else '/dev/sdb' }}"
           volume_type: gp2
-          volume_size: 50
+          volume_size: 70
```
mffiedler: The size of the docker device was previously a concern for qcow2.   We had to keep it down to 20GB or less else it impacted storage in OpenStack.   Is this no longer a concern?  Is 70GB really required?

chaitanyaenr: @jeremyeder ^

jeremyeder: @mffiedler what environment were you using where the size was a problem?  Deathstar had no CoW Ceph to back up the VMs.  Last test we used almost no space because of the CoW, even though each VM was technically 50GB in size (and we had 2k of them).

mffiedler: For AWS, the EBS volumes are not COW.  gp2 is pretty cheap, but we're leaving behind a lot of snapshots currently by folks not understanding how AMIs work.  Not opposed to the size increase, just want to make sure we need it hard-coded in the playbook vs. in the inventory.

chaitanyaenr: @mffiedler I think it's a good idea to parametrize the volume_size, will update the PR.

chaitanyaenr:
@mffiedler updated the PR with the changes requested, PTAL.

image_provisioner/playbooks/roles/aos-ansible/tasks/main.yaml
```diff
@@ -37,3 +37,20 @@
...
+- block:
+   - name: check if skopeo is installed
+     command: which skopeo
+     register: skopeo_status
+
+   - name: Install skopeo if not installed
+     yum:
+       name: skopeo
+       state: latest
+     when: ( not atomic | default(False)| bool and skopeo_status.rc != 0)
```
sjug: do we need the when `( )`?

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+        name: container-selinux
+        state: present
+
+    # Used to pull and install the system container
+    - name: Ensure atomic is installed
+      package:
+        name: atomic
+        state: present
+
+    - name: Ensure runc is installed
```
sjug: Do you have three different package commands because that's what they did upstream?

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+  set_fact:
+    crio_image_prepend: "{{ crio_registry }}"
+    crio_image_name: "cri-o"
+
+- name: Set the full image name
+  set_fact:
+    crio_image: "{{ crio_image_prepend }}/{{ crio_image_name }}:latest"
+
+- name: Pre-pull CRI-O System Container image
+  command: "atomic pull --storage ostree {{ crio_image }}"
```
sjug: Pretty sure we can do this with `atomic_image`

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+    line: "\t\"overlay2.override_kernel_check=1\""
+    state: present
+
+- name: create cni directory to store network settings
+  file:
+    path: /etc/cni/net.d
+    state: directory
+
+- name: create directory to store plugins
+  file:
```
sjug: As above, do we have two create directory steps due to how it was done upstream?

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
chaitanyaenr: Yes, we need atomic package to be installed to pull the crio image and install it.

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
chaitanyaenr: Yes.

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
sjug: Yes, of course we do. I mean do we literally need discrete tasks to install each rather than a loop `with_items`?

chaitanyaenr: You are right, we don't need discrete tasks. Updated the PR with the changes.

sjug:
LGTM

chaitanyaenr:
@mffiedler PTAL.

mffiedler:
@chaitanyaenr sorry, missed your PTAL.
@wabouhamad Can you test our scenario with these changes?  No need to test cri-o - just our normal AMI creation.  Thanks.

jeremyeder:
Does this work?

wabouhamad:
We keep hitting a yum update error when running from this PR branch, even after its been rebased:

Hitting issue described below where $releasever is not resolving to "7Server", but instead to "%24releasever", causing yum update to fail:
https://access.redhat.com/solutions/1417753

Resolution
Added an ansible task before yum update to execute these two commands:
touch /etc/yum/vars/releasever 
echo "7Server" >> /etc/yum/vars/releasever

Got past that error but hitting new errors pulling imaged from reg-aws registry.
Need to retest and troubleshoot the registry pull errors encountered.

chaitanyaenr:
aos-ansible is not able to pull the images even when the crio is set to False when using this branch.

chaitanyaenr:
@vikaslaad PTAL.

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+    regexp: '^cgroup_manager'
+    line: "cgroup_manager = \"systemd\""
+
+- name: Ensure that file locking is set to false
+  lineinfile:
+    path: /etc/crio/crio.conf
+    regexp: '^file_locking'
+    line: 'file_locking = false'
+
+- name: override overlay2 kernel check
```
jeremyeder: Question -- why do we need to mangle crio.conf?  What is wrong with the defaults?

image_provisioner/playbooks/roles/cri-o/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+  lineinfile:
+    path: /etc/crio/crio.conf
+    insertafter: '^storage_option'
+    line: "\t\"overlay2.override_kernel_check=1\""
+    state: present
```
jeremyeder: This was necessary 6 months ago when this PR was written.  Is it still necessary?

image_provisioner/playbooks/roles/cri-o/files/crio-bridge.conf
```diff
@@ -0,0 +1,15 @@
+{
+    "cniVersion": "0.2.0",
+    "name": "crio-bridge",
+    "type": "bridge",
+    "bridge": "cni0",
+    "isGateway": true,
+    "ipMasq": true,
+    "ipam": {
+        "type": "host-local",
+        "subnet": "10.88.0.0/16",
```
jeremyeder: What is this?

PR #303, opened 2017-09-18
Avoid installing atomic-openshift-master package

chaitanyaenr:
The package needs to be installed on just the master in order to
differentiate between master and rest of the nodes. openshift-ansible
will take care of installing it on the master while setting up openshift.

### DIFF
--- a/image_provisioner/playbooks/roles/openshift-package-install/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/openshift-package-install/tasks/main.yaml
@@ -2,7 +2,6 @@
 - name: install openshift packages
   yum: name={{ item }} state=latest
   with_items:
-    - atomic-openshift-master
     - atomic-openshift-sdn-ovs
     - atomic-openshift-tests
     - atomic-openshift-node


mffiedler:
This sort of goes against our approach of having everything pre-loaded.  In rare cases like the original CNCF test were we are isolated we have to get the package somewhere it can be installed.

What problem are you trying to solve with this? Can you check if you are on a master by testing to see if the master systemd units are enabled?

chaitanyaenr:
@jmencak  tuned setup queries for files under /etc/origin/master to know whether it's a master node. I will check with him to see if checking the master systemd units work. Thanks @mffiedler.

jmencak:
Why have master packages on non-master nodes?  The problem is that atomic-openshift-master rpm writes an unpackaged config /etc/origin/master/master-config.yaml which is one of the indications that the node is/should_be a master node.  The new tuned hierarchy checks for the presence of that file.  A check for an enabled systemd is a possibility, I guess, but in my opinion it is better to keep things clean and not install packages which are not strictly required, especially if they are not big packages to download.

jeremyeder:
We put all packages everywhere so that we'd only have one gold image for every node type.

I think it's easier to not design for the CNCF case.  I think having a single image for every possible openshift role outweighs the downside of the masters having to pull RPMs or images.  Those pulls would only occur on 3 machines in the cluster and wouldn't affect the install speed very much.

Sound reasonable?

jmencak:
Yeah, why not close it, Segug can take over the tuned hierarchy project from me.

mbruzek:
This PR is removing atomic-openshift-master from the installed RPM when creating the gold image.

As a consumer (scale-ci automation scripts) of the gold image, I am confused when reading these comments. To install the openshift-master RPM we need the playbooks in the `openshift-ansible` directory which is what the RPM provides. As I understand the image builder scripts will then clone the openshift-ansible github repository in the `/root` directory. So how would a new image know where to get the corresponding atomic-openshift-master RPM of the same version number when the scale-ci script installs. 

@sjug 's last comment suggested closing this PR when I read @jeremyeder 's last comment as we need this PR. Which is it?

jeremyeder:
@mbruzek "To install the atomic-openshift-master RPM we need the playbooks in the openshift-ansible directory which is what the RPM provides."

That's not true, we install the RPMs without ansible here:
https://github.com/openshift/svt/blob/master/image_provisioner/playbooks/roles/openshift-package-install/tasks/main.yaml

I can't see at this point how removing these RPMs will impact scale-ci work.

@sjug not close PR ... merge PR.  Unless you know why it shouldn't be merged.

sjug:
I don't have any other reason to keep it a single master image other than that's how it has been done this far, like @mffiedler said. If you want to move away from that to fix tuned detection, then that's what we'll do.

mbruzek:
@jeremyeder Yeah am still confused at where the VM image gets the actual openstack-ansible playbooks. I thought they were part of the atomic-openshift-master rpm because I found one image without the `/usr/share/ansible/openshift-ansible` directory.

@jeremyeder wrote: 
> "That's not true, we install the RPMs without ansible here:
> https://github.com/openshift/svt/blob/master/image_provisioner/playbooks/roles/openshift-package-install/tasks/main.yaml"
That looks like an Ansible role installing a list of RPMs. That is actually the file that is involved in this pull request. My question was: If we merge this PR and no longer install the "atomic-openshift-master" in this playbook. How will the resulting image get the master code when it needs it?

That issue aside I was simply interested in the location of the Ansible byo playbooks that the openshift-ansible-contrib repo uses in the install process. And concerned if making this change would require another dependency or prerequisite for the scale-ci automation scripts.

Merge or merge not, there is no try.

mffiedler:
@mbruzek I think there is some confusion between installing the OpenShift RPMs and what the playbooks do.   Here are the main OpenShift binary RPMs (forget openshift-ansible for now - that is a red herring)  :

# yum list installed atomic-openshift*
Loaded plugins: amazon-id, rhui-lb, search-disabled-repos
Installed Packages
atomic-openshift.x86_64                                                                                    3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-clients.x86_64                                                                            3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-clients-redistributable.x86_64                                                            3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-docker-excluder.noarch                                                                    3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-dockerregistry.x86_64                                                                     3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-excluder.noarch                                                                           3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-master.x86_64                                                                             3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-node.x86_64                                                                               3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-pod.x86_64                                                                                3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-sdn-ovs.x86_64                                                                            3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-tests.x86_64                                                                              3.7.0-0.127.0.git.0.459b70b.el7                                                             @aos
atomic-openshift-utils.noarch                                                                              3.7.0-0.127.0.git.0.b9941e4.el7                                                             @aos

Installing these RPMs just puts binaries on disk (for the most part).   It does nothing towards configuring an operational cluster.   It gets the binaries out of the repo, across the network (the important bit) and on to disk.

The openshift-ansible playbooks (whether from GitHub or installed via the openshift-ansible RPMs - makes no difference)  do the heavy lifting of installing and configuring a cluster.  One of the steps in the playbooks is to install the above RPMs **if they are not already installed**.   The reason for pre-installing the above binary RPMs on our gold images is to save the network transfer - thousands of transfers of the same binaries is wasteful.

So,  for just 3 masters per cluster it really does not matter whether the master RPMs are preinstalled.   It is a matter of a few extra minutes to let the playbooks install the master RPMs.  As a % of  the total cluster install,  it is negligible.   The nodes are the real burden - pulling all of their binaries (openshift, openvswitch, docker, etc) over and over is a huge time sink. 

Hope this helps some - feel free to ping to continue the discussion.    I think this description is good for a GitHub PR, but more detail should likely be offline.

PR #308, opened 2017-09-26
Adding a delete-oldest role to delete old images.

### LINKED ISSUE #304: The image server disk is full !
The internal image server disk filled up earlier this week (`df` reported the 1.8T disk was 100% used). The image server contains two directories where the images are stored the `rhel` directory currently has 19 30GB image files, while the `atomic` directory has 26 30GB image files. We deleted the oldest images in each directory but we need to come up with an automated way to ensure that never happens again. 

I had a discussion with @chaitanyaenr about how to resolve this. We discussed several options and decided to create an issue to have a broader discussion about the best way to solve this.

## Use a cron job
The first thought was to create a cron job to run a cleanup script on regular intervals.

## Delete n images before Creating n images
The second option was to delete an image(s) before creating new ones. If for example the script that creates images is going to create 2 images, then delete 2 of the oldest images to ensure the disk size stays even.

## Jenkins cleanup job
The third option we discussed was to create a separate Jenkins job to run a cleanup script on regular intervals. 

Please post to this issue how you would solve the problem and we can discuss the pros and cons of each approach.

mbruzek:
This PR is to fix #304 
@chaitanyaenr please review

### DIFF
--- a/image_provisioner/playbooks/delete_oldest.yaml
+++ b/image_provisioner/playbooks/delete_oldest.yaml
@@ -1,5 +1,0 @@
----
-- name: Delete the oldest file in the directory
-  hosts: localhost
-  roles:
-    - delete-oldest
--- a/image_provisioner/playbooks/delete_oldest.yaml
+++ b/image_provisioner/playbooks/delete_oldest.yaml
@@ -1,0 +1,5 @@
+---
+- name: Delete the oldest file in the directory
+  hosts: localhost
+  roles:
+    - delete-oldest
--- a/image_provisioner/playbooks/roles/delete-oldest/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/delete-oldest/tasks/main.yaml
@@ -1,0 +1,74 @@
+---
+# Abort if target_directory is not set.
+- name: Checking for target_directory
+  fail:
+    msg: "Must provide the target_directory variable."
+  when: target_directory is not defined
+
+# Get the disk usage numbers.
+- name: Get the disk usage information before cleanup
+  command: "df --output=avail,pcent {{ target_directory }}"
+  register: disk_space
+
+# Create the disk space facts.
+- name: Create the disk usage facts for {{ target_directory }}
+  set_fact:
+    available_bytes: "{{ disk_space['stdout_lines'][1].split()[0] }}"
+    percent_used: "{{ disk_space['stdout_lines'][1].split()[1] }}"
+
+# Print out the disk usage.
+- name: Print out the disk usage information
+  debug:
+    msg: "There are {{ available_bytes }} bytes available with {{ percent_used }} used on this partition."
+
+# Find the files.
+- name: Find the files in {{ target_directory }}
+  find:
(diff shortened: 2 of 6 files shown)


image_provisioner/playbooks/delete_oldest.yaml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Delete the oldest file in the directory
+  hosts: localhost
+  roles:
```
chaitanyaenr: Instead of a separate playbook, can you call this role from existing playbooks?

image_provisioner/playbooks/roles/delete-oldest/tasks/main.yaml
```diff
@@ -0,0 +1,74 @@
+---
+# Abort if target_directory is not set.
+- name: Checking for target_directory
+  fail:
+    msg: "Must provide the target_directory variable."
+  when: target_directory is not defined
```
chaitanyaenr: can you add the target_directory variable to the sample inventory files?

image_provisioner/playbooks/delete_oldest.yaml
chaitanyaenr: Also can you make this role optional?

sjug:
Is deletion of a single image the desired functionality?

image_provisioner/playbooks/delete_oldest.yaml
mbruzek: @ravi yes we can do that. I will add it to the provision_gold_images.yaml playbook.

image_provisioner/playbooks/roles/delete-oldest/tasks/main.yaml
mbruzek: I can add the variable to the files, but the way our Jenkins calls it is by copying an inventory outside of this repository so I will also add it to the inventory that Jenkins uses.

mbruzek:
This latest commit addresses the review comments from @chaitanyaenr and @sjug. I added the variable to the inventory files and tied it to `provision_gold_images.yaml` rather than having a separate playbook.

@sjug re: desired functionality. We ran out of disk space on the image server (several times) and came up with a solution to delete one image when attempting to create one image to avoid filling up the disk in the future.

image_provisioner/playbooks/roles/delete-oldest/tasks/main.yaml
```diff
@@ -0,0 +1,74 @@
+---
+# Abort if target_directory is not set.
+- name: Check for the target_directory variable
```
chaitanyaenr: I don't think we need this check since the role is run only when the target_directory variable is set.

mbruzek: Since that check is in a different file I would like to keep the check here. If anyone tries to call this role without that variable set, they will know exactly what is wrong and what to fix.

chaitanyaenr:
LGTM

PR #310, opened 2017-09-27
Optimize journald configuration for system and scalability testing

mffiedler:
1. Disable rsyslog
2. Enable journal persistence on disk instead of /run/systemd/journal
3. Decrease rate limiting to 10K msgs/sec

@wabouhamad @chaitanyaenr  @sjug PTAL

Tested with the AMI playbook but qcow2 should be no different.  @ravisantoshgudimetla made these same updates to the Atomic image successfully.

### DIFF
--- a/image_provisioner/playbooks/roles/journald-config/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/journald-config/tasks/main.yaml
@@ -1,0 +1,56 @@
+---
+- name: Stop and disable rsyslog
+  systemd:
+    name: "{{ item }}"
+    state: stopped
+    enabled: no
+  with_items:
+    - rsyslog
+  when: not atomic | default(False) | bool
+
+- name: Enable journald persistence
+  ini_file:
+    dest: "/etc/systemd/journald.conf"
+    section: Journal
+    option: Storage
+    value: "Persistent"
+    no_extra_spaces: yes
+
+- name: Increase journald rate limit burst to 10000
+  ini_file:
+    dest: "/etc/systemd/journald.conf"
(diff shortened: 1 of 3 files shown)


ravisantoshgudimetla:
cc @chaitanyaenr

jeremyeder:
@mffiedler @portante shouldn't this be considered a merge into openshift-ansible itself?

mffiedler:
Ooops - sorry, Ravi G ( @ravisantoshgudimetla )

mffiedler:
@jeremyeder IMO this is stuff in a sysadmin's discretion.   I can open a doc bz for recommendations, but things like disabling rsyslog and tuning the burst settings have been batted back and forth in terms of actually setting them in openshift-ansible.  example:. https://bugzilla.redhat.com/show_bug.cgi?id=1380458

wabouhamad:
@mffiedler, LGTM

portante:
@jeremyeder, do you mean having a PR like this one?  https://github.com/openshift/openshift-ansible/pull/3753/files

image_provisioner/playbooks/roles/journald-config/tasks/main.yaml
```diff
@@ -0,0 +1,55 @@
+---
+- name: Stop and disable rsyslog
```
chaitanyaenr: rsyslog is not installed by default on atomic, so we need to add a condition to disable it only when not atomic. Except that LGTM.

mffiedler:
Added a conditional for disabling rsyslog

sjug:
LGTM

PR #312, opened 2017-10-06
Updated os-kickstart role to yum update redhat-release-server optiona…

wabouhamad:
…lly, and in role seal-image deleting /root/.kube dir and emptying /etc/origin/node and /etc/origin/master directories

### DIFF
--- a/image_provisioner/playbooks/roles/seal-image/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/seal-image/tasks/main.yaml
@@ -3,6 +3,14 @@
   shell:  yum clean all
   ignore_errors: true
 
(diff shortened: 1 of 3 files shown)


image_provisioner/playbooks/roles/seal-image/tasks/main.yaml
```diff
@@ -3,6 +3,14 @@
   shell:  yum clean all
   ignore_errors: true
 
+- name: remove /root/.kube dir
+  shell: /bin/rm -rf /root/.kube
```
mffiedler: Probably preferable to use the file task with state=empty.   Future enhancement.

PR #313, opened 2017-10-06
Remove /etc/origin before sealing image

mffiedler:
Remove /etc/origin and remove on-disk yum cache before sealing image.

@wabouhamad PTAL


wabouhamad:
@mffiedler LGTM

PR #317, opened 2017-10-09
Bake in pbench agent and controller images even when not atomic

chaitanyaenr:
For scale tests, jump host and cns nodes are RHEL based, so we need 
to bake in pbench images even not atomic.

### DIFF
--- a/image_provisioner/playbooks/roles/pbench-kickstart/tasks/main.yml
+++ b/image_provisioner/playbooks/roles/pbench-kickstart/tasks/main.yml
@@ -1,19 +1,21 @@
 ---
-- name: create directory to store pbench results, store openshift templates for pbench
-  file: 
-    path: "{{ item }}"
-    state: directory
-  with_items:
-     - /var/lib/pbench-agent
-     - /var/home/openshift-templates/pbench
+- block:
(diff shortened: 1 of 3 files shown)


chaitanyaenr:
@openshift/svt

chaitanyaenr:
@sjug PTAL.

sjug:
Small change LGTM, thanks @chaitanyaenr

chaitanyaenr:
@sjug Thanks for the review. @mffiedler PTAL.

image_provisioner/playbooks/roles/pbench-kickstart/tasks/main.yml
```diff
@@ -1,19 +1,21 @@
...
-    owner: root 
-    group: root 
-    mode: 0644
+    - name: copy openshift template for pbench
+      copy: 
+        src: pbench-agent-daemonset.yml  
+        dest: /var/home/openshift-templates/pbench 
+        owner: root 
+        group: root 
+        mode: 0644i
```
sjug: Eh I think this might be a typo?

chaitanyaenr: @sjug good catch, will fix it in one of the existing PR.

PR #320, opened 2017-10-19
Set json-file as default logging driver

chaitanyaenr:


chaitanyaenr:
@mffiedler PTAL.

mffiedler:
LGTM

PR #322, opened 2017-10-24
Initial commit of nvidia and stac-a2 content

jeremyeder:
This is largely a WIP as we work with NVIDIA to streamline many things uncovered during these exercises.  I do not know if this branch will ever be updated.  Fair warning :smile: 

The reason there are systemd files in here is because the RHEL7.4 base container image GA included a systemd errata that was not included in the RHEL7.4 GA.  When a system is properly registered or attached to a Satellite, none of this is necessary.  However our benchmark system only had access to RHEL7.4 GA content, so this was necessary.

I am committing this just so that folks watching can use these Dockerfiles as-is, even though in a properly configured/registered system that had access to normal content/errata from Red Hat, this issue would not have come up.


jeremyeder:
@vikaschoudhary16 I'm going to merge for now, we can merge any feedback/fixups from your POV when you return.

vikaschoudhary16:
@jeremyeder Should we also add notes in README related to nvidia-docker2 installation issue? For example it should be installed with --no-deps or info about the repos that flx42 shared for docker and nvidia-container-runtime?

jeremyeder:
For the flx42 part, he merged the fixes into NVIDIA's production yum repository.
For the nvidia-docker2, I am pursuing that with them offline for the real fix to their rpm spec file.

I will add a note about --no-deps for now...that should be fixed within a couple weeks.

PR #327, opened 2017-10-31
Add tags to qcow images

chaitanyaenr:
The qcow images needs to be tagged in order to differentiate between
rhel and atomic gold images.


chaitanyaenr:
@openshift/svt

sjug:
Small change, lgtm.

mbruzek:
@chaitanyaenr This impacts the scale-ci work, as it uses regular expressions to find the images. What is the reason for this change.

PR #334, opened 2017-11-08
Prevent yum update failures when not updating to RHEL 7.5

wabouhamad:
This was needed to prevent yum update failure when trying to update pkg initscripts when redhat-release-server is not updated to higher version 7.5 beta. Modified role os-kickstart task yum update to exclude pkg initscripts and only update it when pkg redhat-release-server gets updated

### DIFF
--- a/image_provisioner/playbooks/roles/os-kickstart/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/os-kickstart/tasks/main.yaml
@@ -5,12 +5,15 @@
     yum: 
       name: '*' 
       state: latest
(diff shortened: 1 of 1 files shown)


wabouhamad:
@mffiedler @vikaslaad @chaitanyaenr @sejug,  PTAL

wabouhamad:
@sjug PTAL

chaitanyaenr:
LGTM

sjug:
Haha, @wabouhamad I think you got the title and body mixed up a bit there.
Very minor change, lgtm.

PR #337, opened 2017-11-13
Add templates to deploy tooling on OpenShift

chaitanyaenr:
This commit adds:
- pbench-agent and collectd templates.
- a readme with instructions about deploying the tooling.


chaitanyaenr:
@openshift/svt

openshift_templates/collectd-config.yml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: collectd-config
+data:
+  host: " "
```
mbruzek: Why are these one space? Could they be empty string?

chaitanyaenr: Yes, they can be a empty string. Fixed.

mbruzek:
The readme is well formed markdown and makes sense. The templates look fine to me.
/lgtm

chaitanyaenr:
@openshift/svt ready for merge?

mffiedler:
@chaitanyaenr @mbruzek If we're going to create a new toplevel folder in the repo I'd prefer to create subfolders to organize the content underneath.   Otherwise it will just become a dumping ground.   Please add an additional level of dir structure for your content - maybe "performance_monitoring" or "data_collection"?   

That way we can put other templates not related in other subdirs.   What do you think?

chaitanyaenr:
@mffiedler makes sense, I have moved the templates to subdirs.

PR #339, opened 2017-11-15
Add tool to replace parameters in clusterloader configs

chaitanyaenr:
This commit adds a tool to replace parameters in clusterloader configs
used by http, mastervertical and nodevertical tests in order to avoid
maintaining multiple configs for each test and enable running tests
in CI.

### DIFF
--- a/openshift_scalability/config-replacer/README.md
+++ b/openshift_scalability/config-replacer/README.md
@@ -1,0 +1,69 @@
+# Cluster-Loader config replacer
+Replaces parameters in configs used by http, nodevertical and mastervertical tests.
+
+## Run
+```
+$ python replace.py <test> --config=<path to config> <parameters to be replaced>
+```
+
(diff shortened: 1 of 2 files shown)


chaitanyaenr:
@jmencak added a functionality in golang cluster loader which will look for environment variables and set them in the config. This script can be used when using python cluster loader.

jmencak:
Ravi, while I think this is really useful and I'd use it if I had to use the old Python cluster loader (provided I was unable to change its code to read the environment for parameters), there will probably be no need for this tool when we start using the Golang cluster loader.  And we should start using it now!  I really believe we should try and put more effort into the Golang cluster loader now or at least preparing templates/configs for it.

chaitanyaenr:
@jmencak makes sense to me.

mffiedler:
Closing this PR in favor of enhancements to the golang cluster loader per comments from @jmencak and @chaitanyaenr

PR #345, opened 2017-11-28
Modify template to work with kubernetes 1.7

chaitanyaenr:
In 1.7 release the shared PID namespaces is enabled by default. This is
causing a problem when running pbench container which runs systemd.

Related issue in upstream kubernetes:
https://github.com/kubernetes/kubernetes/issues/48937

According to my understanding this will be fixed from 1.8 with the
following patch:
https://github.com/kubernetes/kubernetes/pull/51634


chaitanyaenr:
@openshift/svt PTAL.

mffiedler:
I don't think that first linked k8s issue is correct - that is some GlusterFS issue

chaitanyaenr:
The problem and the solution were discussed in that issue.

PR #347, opened 2017-11-29
Add templates to create namespaces

chaitanyaenr:
This commit adds templates to create pbench and collectd namespaces with an
empty nodeselector like what we have for default namespace. This will avoid
hitting the following bug: https://bugzilla.redhat.com/show_bug.cgi?id=1494709.


chaitanyaenr:
We used to manually edit the pbench namespace to add the nodeselector before. This is needed to run stuff in ci.

PR #348, opened 2017-11-30
Add support to run mastervert using golang clusterloader

chaitanyaenr:
This commit adds the config and modifies the script to support
running master-vertical using golang clusterloader.

### DIFF
--- a/openshift_scalability/masterVertical.sh
+++ b/openshift_scalability/masterVertical.sh
@@ -5,11 +5,39 @@
 # pbench-user-benchmark -C testname -- ./mastervirt.sh
 # pbench-move-results
 
+if [ "$#" -ne 1 ]; then
+  echo "syntax: $0 <TYPE>"
+  echo "<TYPE> should be either golang or python"
+  exit 1
+fi
+
+TYPE=$1
+
+golang_clusterloader() {
+  # Export kube config
+  export KUBECONFIG=${KUBECONFIG-$HOME/.kube/config}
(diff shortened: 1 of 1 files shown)


mffiedler:
We need to determine what "master vertical" is going forward.  The test we have run release to release up through OCP 3.7 is pyconfigMasterVertScalePause.yaml (or pyconfigMasterVertScale, the hello-openshift variant), not the kitchen-sink type test which is proposed in this PR.  

The current test creates many more resources and resource types per project than the kitchen sink template and I personally prefer it.   Kitchen sink drags along a database with every application and is difficult to use as a control plane density test.  It does not include the secrets, env vars, routes,  etc.

The correct test is probably something in between but we'll need some agreement.   

Opinions?

/cc: @jeremyeder @jmencak @sjug

jmencak:
Given we have a good set of templates already, I'd (re)use the templates from the Python cluster-loader version.  I always assumed the golang cluster loader was going to be at least mostly compatible with the python version, however, it seems this is not the case at the moment (at least for templates) and needs to be fixed first.

chaitanyaenr:
@mffiedler now that we have the config's for golang clusterloader committed, I guess we can merge this right?

mffiedler:
It's fine for now, but we still need a final definition of cluster horizontal and master vertical.  That is my to-do and I'll open an issue for it.   Needs to be resolved in 3.10

chaitanyaenr:
@mffiedler thanks for reviewing the PR.

PR #350, opened 2017-12-01
Add script to setup tooling on openshift

chaitanyaenr:
This commit adds a script which sets up tooling and checks if the
pods are in running state. This will be run by the jenkins job once
it receives a signal and parameters from the scale ci job.

### DIFF
--- a/ci/tooling/setup_tooling.sh
+++ b/ci/tooling/setup_tooling.sh
@@ -1,0 +1,108 @@
+#!/bin/bash
+
+# Check for kubeconfig
+if [[ ! -s $HOME/.kube/config ]]; then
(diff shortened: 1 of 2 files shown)


mffiedler:
In general, I think packages should either be installed via:
    - gold image creation (can we assume this will run on instances created from our images?)
    - package manager (yum, dnf, apt-get) as opposed to wget or curl.

chaitanyaenr:
@mffiedler you are right, I will bake in the dependencies into the gold image.

chaitanyaenr:
@mffiedler made the changes requested, jq will be baked into the gold image but I have left the jq installation steps in there just in case we want to use it on an instance not created using the gold image.

PR #352, opened 2017-12-07
Bake in openshift-ansible code tied to a particular openshift version

chaitanyaenr:
This commit will enable us to use the stable openshift-ansible code
tied to a particular release instead of using the code from the master
branch in order to avoid using the commits made to support the upcoming
release.

### DIFF
--- a/image_provisioner/playbooks/roles/clone-repos/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/clone-repos/tasks/main.yaml
@@ -29,13 +29,26 @@
     path: /root/openshift-ansible 
   register: openshift_ansible
 
-- name: clone openshift-ansible repo
-  git: 
-    repo: https://@github.com/openshift/openshift-ansible 
-    dest: /root/openshift-ansible 
-    update: yes
+- block:
+   - name: set openshift-ansible branch to be cloned
+     set_fact:
+       openshift_ansible_branch: release-{{ openshift_version.split('.')[0] }}.{{ openshift_version.split('.')[1] }}
+
+   - name: clone openshift-ansible repo
+     git: 
+       repo: https://@github.com/openshift/openshift-ansible 
+       dest: /root/openshift-ansible 
+       update: yes
+       version: "{{ openshift_ansible_branch }}"
+  rescue:
+   - name: clone master branch of openshift-ansible
+     git:
+       repo: https://@github.com/openshift/openshift-ansible
+       dest: /root/openshift-ansible
+       update: yes
+       version: master
   when: openshift_ansible.stat.exists == False
-  
+
 #- name: clone kubernetes repo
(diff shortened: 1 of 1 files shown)


chaitanyaenr:
@mbruzek @openshift/svt PTAL.

image_provisioner/playbooks/roles/clone-repos/tasks/main.yaml
```diff
@@ -29,13 +29,22 @@
     path: /root/openshift-ansible 
   register: openshift_ansible
 
+- name: get the openshift version
+  shell: echo "{{ openshift_version }}" |  awk -F '.' '{ print $1"."$2 }'
```
mbruzek: Instead of doing a shell command, we should use Ansible to split the string:  
`openshift_version.split('.')

image_provisioner/playbooks/roles/clone-repos/tasks/main.yaml
```diff
@@ -29,13 +29,22 @@
...
+  register: version
+
+- name: set openshift-ansible branch to be cloned
+  set_fact:
+     openshift_ansible_branch: release-{{ version.stdout }}
```
mbruzek: `"{{ openshift_version.split('.')[0] }}.{{ openshift_version.split('.')[1] }}"` this will avoid having to add an additional shell task just to manipulate the string.

chaitanyaenr:
@mbruzek thanks for the review, I have the updated the PR to use split like you mentioned. PTAL.

mbruzek:
+1 LGTM

chaitanyaenr:
@mffiedler PTAL.

mffiedler:
This change will work for 3.6 and higher.  Is that sufficient?  I assume so.  Previous releases have openshift-ansible branches of release-1.5, release-1.4, etc.

mbruzek:
@mffiedler I requested this change because the master branch currently points to release 3.8 bits and that does not work for 3.7 images. The list of branches can be found here: https://github.com/openshift/openshift-ansible/branches/active 

It looks to me that this fix would be sufficient to cover the most current branches, but if you notice something that is not covered please let us know.

chaitanyaenr:
@mffiedler I think it's sufficient since we won't be using the release branches older than 3.6. Right?

chaitanyaenr:
@mffiedler PTAL.

mffiedler:
@chaitanyaenr I didn't try to run this myself yet.   What does it do for something like 3.9?   There is no release-3.9 branch yet for openshift-ansible - 3.9 == master right now.   If it is not happening already, we should use master as the backup branch.

chaitanyaenr:
@mffiedler I have added a rescue block which will clone the master branch in case it fails to find the appropriate release branch. PTAL.

Issue #357, opened 2017-12-19
Debug prometheus conformance test hang


mffiedler:
Currently, the prometheus conformance test hangs.  Need to debug it and open a bz as appropriate.

extended.test --ginkgo.v --ginkgo.focus="Prometheus"

PR #358, opened 2017-12-19
Temporarily skip hanging Prometheus test - needs debug

mffiedler:
The prometheus conformance test currently hangs and eventually times out.   Skip this test temporarily until it can be debugged (issue #357)


mffiedler:
/cc: @mbruzek You'll want this PR too

PR #364, opened 2018-01-16
updates to SVT CI automation scripts and config files

wabouhamad:
Updates include:
- substituting "oc adm" instead of "oadm" in App Scale Up Down CI playbook
- Reducing the number of projects from 5 to 4 for kitchen-sink test as 5 was overwhelming the nodes even back on 3.7.
- returning the default number of pods from 1000 back to 500 so we can run the Node Vertical tests from CI with 250 pod per node on 2 compute nodes (we do not officially support 500 pods per node)
- pulling .kube/config from master's /etc/origin/master/admin.kubeconfig in node vertical shell script

### DIFF
--- a/openshift_scalability/ci/contents/kitchen-sink-ci.yaml
+++ b/openshift_scalability/ci/contents/kitchen-sink-ci.yaml
@@ -1,47 +1,47 @@
 projects:
-  - num: 5
+  - num: 4
     basename: cakephp-mysql
     tuning: default
     templates:
       - num: 1
         file: ./content/quickstarts/cakephp/cakephp-mysql-pv.json
 
-  - num: 5
+  - num: 4
     basename: dancer-mysql
     tuning: default
     templates:
       - num: 1
         file: ./content/quickstarts/dancer/dancer-mysql-pv.json
 
-  - num: 5
+  - num: 4
     basename: django-postgresql
     tuning: default
     templates:
       - num: 1
         file: ./content/quickstarts/django/django-postgresql-pv.json
 
-  - num: 5
+  - num: 4
     basename: nodejs-mongodb
     tuning: default
     templates:
       - num: 1
         file: ./content/quickstarts/nodejs/nodejs-mongodb-pv.json
 
-  - num: 5
+  - num: 4
     basename: rails-postgresql
     tuning: default
     templates:
       - num: 1
         file: ./content/quickstarts/rails/rails-postgresql-pv.json
 
-  - num: 5
+  - num: 4
     basename: eap64-mysql
     tuning: default
     templates:
       - num: 1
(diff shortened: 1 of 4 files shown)


openshift_scalability/ci/contents/kitchen-sink-ci.yaml
```diff
@@ -1,47 +1,47 @@
 projects:
-  - num: 5
+  - num: 4
```
jeremyeder: what does updating to 3.9 have to do with reducing the # of templates?

openshift_scalability/config/nodeVertical.yaml
```diff
@@ -4,7 +4,7 @@ projects:
     ifexists: delete
     tuning: default
     pods:
-      - total: 1000
+      - total: 500
```
jeremyeder: also here -- why is this being done?  the title of the PR is confusing

openshift_scalability/ci/contents/kitchen-sink-ci.yaml
wabouhamad: This config file is based on svt/openshift_scalability/master-vert-pv.yaml which had 1 project per app.  We started with 5 in the CI Jenkins Test then scaled back to 4 back in April 2017 during OCP 3.5.x when we realized we were overwhelming the nodes with 5.  Updating that value in svt repo so we can run this test on 3.9 and compare to previous results.

openshift_scalability/config/nodeVertical.yaml
wabouhamad: We are returning the default number of pods from 1000 back to 500 so we can run the Node Vertical tests from CI at 250 pods per node and do some NV comparisons of 3.9 vs previous releases.   Was trying to use the same NV config file for CI.  If we want to keep that number at 1000, I could create another config file with 500 and keep it in svt/openshift_scalability/ci/contents/ dir.

wabouhamad: @jeremyeder @mffiedler PTAL

openshift_scalability/ci/contents/kitchen-sink-ci.yaml
wabouhamad: @jeremyeder @mffiedler PTAL

wabouhamad:
@mffiedler recent results from past couple days show 3.9 able to handle 500 pause-pods per node reliably.  Should we have for CI purposes both a 250 and 500 ppn config files, so we can run them both from CI and just specify the config file ?  we could store in svt/openshift_scalability/ci/contents/ dir

mffiedler:
For CI purposes it really only makes sense to run the higher number - let's just leave it 500/node for now if that is reliable.   Sorry for the churn.

PR #368, opened 2018-01-17
[WIP] Introduce benchmark framework

mrsiano:
based on the A B test work for pti spectre node vertical benchmark, adding a script that covers that use case for the next future.

also we can abstract it, and having a scrip for any A B test, see TODO inline.

### DIFF
--- a/openshift_scalability/scripts/ptibenchmark.sh
+++ b/openshift_scalability/scripts/ptibenchmark.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# The following script will execture A B Benchmark for pti(spectre).
+
+type=$1
+aname=$2
+bname=$3
+cd `dirname $0` # Make sure to be at svt dir
+
+#TODO: Abstract more nicely.
+
+function usage() {
+echo "
+USAGE:
+    nohup ./ptibenchmark.sh python ptienabled ptidisabled <script> <scriptargs> 2>&1 &
+"
+}
+
+#TODO: handle empty args, defaults.
+
+#A Test
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo 1 > /sys/kernel/debug/x86/pti_enabled; echo 1 > /sys/kernel/debug/x86/ibpb_enabled; echo 1 > /sys/kernel/debug/x86/ibrs_enabled"; done
+pbench-user-benchmark .././nodeVertical.sh $aname $type > /tmp/n.log
+pbench-move-results --prefix=$aname
+
+#B Test
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo 0 > /sys/kernel/debug/x86/pti_enabled; echo 0 > /sys/kernel/debug/x86/ibpb_enabled; echo 0 > /sys/kernel/debug/x86/ibrs_enabled"; done
+pbench-user-benchmark .././nodeVertical.sh $bname $type > /tmp/n.log
+pbench-move-results --prefix=$bname
+
+exit 0


mrsiano:
@jeremyeder @wabouhamad @mffiedler PTAL

openshift_scalability/scripts/ptibenchmark.sh
```diff
@@ -0,0 +1,17 @@
...
+#Optimize the register script to run this command there.
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo "export LC_CTYPE=en_US.UTF-8 >> ~/.bashrc; export LC_ALL=en_US.UTF-8 >> ~/.bashrc; pbench-register-tool-set; done
+
+#PTI Enabled
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo 1 > /sys/kernel/debug/x86/pti_enabled"; done 
+nohup $(pbench-user-benchmark ~/svt/openshift_scalability/./nodeVertical.sh ptibench python > /tmp/n.log 2>&1)  &
+pbench-move-results --prefix=pti_enabled
+
+#PTI Disabled
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo 0 > /sys/kernel/debug/x86/pti_disabled"; done
```
mffiedler: kernel parameter is pti_enabled, not pti_disabled

mffiedler:
My feeling is this belongs in the Jenkins job, not as a standalone script in SVT, but maybe I'm quibbling.

openshift_scalability/scripts/ptibenchmark.sh
mrsiano: right, this will make sure \ force we will test both cases

mrsiano:
Ok, make sense, for now I just represents who we should approach it, we can move it tough.

jeremyeder:
Few things:

 - That one knob doesn't fully disable the security, and most importantly it does not disable the "big impact" performance impacting one (ibrs_enabled).  You're missing two additional sysfs knobs.  https://access.redhat.com/articles/3311301
 - What is the UTF8 stuff about?  I've never seen that be a problem before.
 - I think this should probably be a tuned profile.
 - Last and most importantly ... we can't currently test the impact of these patches on EC2.  It is a long story but in short, Xen-based instances on EC2 are insecure right now.  We have to wait for EC2 to patch their hypervisor.

mrsiano:
1. tuned sounds good idea!
2. about I'v got perl errors in promt so this thing just fixed that, just need to move it aside.

openshift_scalability/scripts/ptibenchmark.sh
```diff
@@ -0,0 +1,20 @@
...
+#Optimize the register script to run this command there.
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo "export LC_CTYPE=en_US.UTF-8 >> ~/.bashrc; export LC_ALL=en_US.UTF-8 >> ~/.bashrc; pbench-register-tool-set; done
+
+#PTI Enabled
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo 1 > /sys/kernel/debug/x86/pti_enabled"; done 
+pbench-user-benchmark ~/svt/openshift_scalability/./nodeVertical.sh ptibench python > /tmp/n.log
+pbench-move-results --prefix=pti_enabled
+
+#PTI Disabled
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo 0 > /sys/kernel/debug/x86/pti_disabled"; done
```
mffiedler: To disable the patch:   echo 0 > /sys/kernel/debug/x86/pti_enabled   See:  https://access.redhat.com/articles/3311301

openshift_scalability/scripts/ptibenchmark.sh
```diff
@@ -0,0 +1,20 @@
...
+#	nohup ./ptibenchmark.sh 2>&1 &
+
+#TODO: Abstract this script to have an A B node vertical concept, but do some changes in between the runs.
+
+#Optimize the register script to run this command there.
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo "export LC_CTYPE=en_US.UTF-8 >> ~/.bashrc; export LC_ALL=en_US.UTF-8 >> ~/.bashrc; pbench-register-tool-set; done
+
+#PTI Enabled
+for h in `oc get no |grep -v NAME|awk '{print $1}'`; do ssh $h "echo 1 > /sys/kernel/debug/x86/pti_enabled"; done 
+pbench-user-benchmark ~/svt/openshift_scalability/./nodeVertical.sh ptibench python > /tmp/n.log
```
wabouhamad: Can we pass the "python" argument for nodeVertical.sh script as command line arg to the ptibenchmark.sh script, so we can run either "python" or "golang" version of cluster loader

openshift_scalability/scripts/ptibenchmark.sh
mrsiano: seems like to completely disable it we need 3 of them to be 0 pti, ibpb, ibrs
will change that

openshift_scalability/scripts/ptibenchmark.sh
mrsiano: sure, we'll abstract it more.

mrsiano:
@jeremyeder
If we go with tuned approach, we need something to control the profiles.

mffiedler:
Closing inactive PR.

PR #371, opened 2018-01-23
Rename the image in the pod spec

chaitanyaenr:
openshift 3.9 is looking for library/pbench-agent on the nodes instead
of just pbench-agent. This commit renames the image in the template and
tags the image accordingly while building the gold image.


chaitanyaenr:
@openshift/svt PTAL.
/cc @jeremyeder

chaitanyaenr:
@mffiedler PTAL.

mffiedler:
@chaitanyaenr double checking that this is tested?

chaitanyaenr:
@mffiedler I have tested it but I am still not sure about why it's looking for library/pbench-agent

PR #374, opened 2018-01-25
Ignore errors when cloning a particular openshift-ansible branch

chaitanyaenr:
This commit will ensure that the playbook doesn't have a failure in 
the summary when it doesn't find a particular openshift-ansible 
release branch and clones the master branch instead.


chaitanyaenr:
@mffiedler PTAL.

wabouhamad:
LGTM

PR #377, opened 2018-02-02
Update selinux-policy

chaitanyaenr:
This commit updates selinux-policy needed for ocp 3.9 installation


chaitanyaenr:
@mffiedler @jmencak PTAL.

jeremyeder:
For some background, I think we were blacklisting those packages because they were broken at the time.  Not aware of any other reason.

chaitanyaenr:
@jeremyeder you are right, that was the only reason.

mbruzek:
+1

chaitanyaenr:
@mffiedler PTAL.

mffiedler:
@wabouhamad Please review

PR #378, opened 2018-02-06
SVT conformance wrapper updates

mffiedler:
In 3.9 some updates to the wrapper are required to get things to run cleanly:

1. Fix Jenkins image stream to latest
2. Remove defaultNodeSelector prior to running tests and restore it afterwards
3. Return a meaningful exit code.

There is still 1 failure right now that looks like a test bug - the test panics.  Need to debug it.

Tested on HA and single master

### DIFF
--- a/conformance/removeNodeSelector.yaml
+++ b/conformance/removeNodeSelector.yaml
@@ -1,0 +1,21 @@
+---
+- hosts: all
+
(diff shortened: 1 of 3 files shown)


mffiedler:
@wabouhamad please review

PR #392, opened 2018-02-23
Added kernel-devel to list of packages to yum install in os-kickstart…

wabouhamad:
Added package "kernel-devel" to list of packages to yum install in os-kickstart role


wabouhamad:
@mffiedler PTAL

PR #396, opened 2018-03-02
Updated task install openshift packages ito install latest openvswitch from rhel7-fast-datapath-rpms repo

wabouhamad:
Installing atomic-openshift-sdn-ovs pkg will also install as a dependency the openvswitch pkg.
In order to install the latest openvswitch pkg from the staging rhel7-fast-datapath-rpms repo, we needed to disable the rhel7-fast-datapath-htb repo.

### DIFF
--- a/image_provisioner/playbooks/roles/openshift-package-install/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/openshift-package-install/tasks/main.yaml
@@ -1,15 +1,9 @@
 ---
-## pre-installing the openvswitch version 2.9.0 so it won't pick up  
-## version "1:2.8.0-4.el7fdb" from rhel7-fast-datapath-htb  
-- name: install openvswitch-2.9.0 from rhel7-fast-datapath-rpms repo
-  yum:
-    name: openvswitch-2.9.0
-    disablerepo: "*"
-    enablerepo: rhel7-fast-datapath
-    state: present
-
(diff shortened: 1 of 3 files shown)


wabouhamad:
@mffiedler PTAL

image_provisioner/playbooks/roles/openshift-package-install/tasks/main.yaml
```diff
@@ -1,6 +1,9 @@
 ---
-- name: install openshift packages
-  yum: name={{ item }} state=latest
+- name: install openshift packages with openvswitch from rhel7-fast-datapath-rpms by disabling rhel7-fast-datapath-htb repo 
+  yum: 
+    name: "{{ item }}"
+    state: latest
+    disablerepo: rhel7-fast-datapath-htb
```
mffiedler: My concern is that we could still lose the package via yum update later (at runtime, not during provisioning).   But I guess that is out of our control since aos-ansible is mucking with the same repos.

wabouhamad: We would need to add the new staging repo Justin is sync-ing from now, as a task in the repo-install role

mffiedler:
We will eventually revert this once OVS 2.9 is available in GA channels, but merging for now since we will need this until OVS 2.9 GAs

PR #406, opened 2018-03-16
Add extra_tags option for AMI's

chaitanyaenr:
This commit allows us to add extra tags to the gold image name. This
is needed to differentiate m5.x supported images from others.


chaitanyaenr:
@wabouhamad PTAL.

PR #416, opened 2018-04-03
Script to configure taints on nodes, run Golang cluster-loader and then remove taints after test

wabouhamad:
This script will:

1. configure taints on computes nodes
2. run Golang cluster-loader with taints and tolerations config file: node-taints-pod-tolerations.yaml which is already in svt repo:  openshift_scalability/config/golang dir
3. delete the projects created by cluster-loader after all the pods with tolerations have been deployed
4. remove taints from compute nodes


wabouhamad:
@mffiedler  PTAL

PR #417, opened 2018-04-03
Script to do node configurations then run Golang cluster-loader for Node Affinity Anti-Affinity scheduling test

wabouhamad:
This script will:

1. label the compute nodes
2. execute Golang cluster-loader with config file node-affinity-anti-affinity.yaml already in svt repo
3. clean up projects and remove labels from nodes after test


wabouhamad:
@mffiedler PTAL

PR #423, opened 2018-04-12
Script to start pbench and run golang cluster loader for SVT Create Projects Loaded with Artifacts testcase

wabouhamad:
Script in svt/openshift_scalability/ci/scripts:
- pbench-start-create-loaded-projects.sh.  Starts pbench on pbench-controller test client, then runs golang cluster loader with config yaml file.  Stops pbench, postprocesses results and copies results to pbench server.  Provides links to pbench results 

Golang Cluster loader config file based on pyconfigLoadedProject.yaml:
- configCreateLoadedProjects.yaml


wabouhamad:
@vikaslaad PTAL

PR #428, opened 2018-04-16
Replace atomic-openshift-utils package with atomic-openshift

chaitanyaenr:


chaitanyaenr:
@wabouhamad @mffiedler PTAL.
/cc @jeremyeder

wabouhamad:
LGTM

PR #429, opened 2018-04-17
Switch to pbench namespace after creating it using the template

### LINKED ISSUE #424: setup_pbench_pods.sh doesn't create pbench-agent pods in pbench namespace
https://github.com/openshift/svt/blob/master/openshift_tooling/pbench/setup_pbench_pods.sh#L87

@chaitanyaenr , PTAL

chaitanyaenr:
Fixes #424

### DIFF
--- a/openshift_tooling/pbench/setup_pbench_pods.sh
+++ b/openshift_tooling/pbench/setup_pbench_pods.sh
@@ -66,9 +66,9 @@ function short_sleep() {
 # cleanup deamonset if already exists
 function cleanup() {
 	# pbench cleanup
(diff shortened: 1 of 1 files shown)


chaitanyaenr:
@jmencak PTAL.

jmencak:
LGTM, however, you may want to use '-n pbench' when crreating resources, just in case someone/something doesn't change the namespaces while the script is running.

mffiedler:
+1 to @jmencak suggestion

chaitanyaenr:
@jmencak @mffiedler makes sense, updated the PR with the changes requested. PTAL.

PR #430, opened 2018-04-17
ovs and atomic-openshift-sdn now run as a daemonset.  Do not install.

mffiedler:
@wabouhamad @chaitanyaenr PTAL

This is a breaking change for generating images for older releases of OpenShift.  Is it time to have release branches of svt (like openshift-ansible does)?   Thoughts?

In the interim, we need to disable openvswitch on our clusters so that the ovs pods don't crash loop

### DIFF
--- a/image_provisioner/playbooks/roles/repo-install/templates/etc/yum.repos.d/all.repo.j2
+++ b/image_provisioner/playbooks/roles/repo-install/templates/etc/yum.repos.d/all.repo.j2
@@ -53,15 +53,6 @@ enabled=1
 gpgcheck=0
 sslverify=0
 
-# This htb repo gets us openvswitch-2.6, and is required until the htb bits are promoted to the above GA channel (GA is openvswitch-2.5 for now).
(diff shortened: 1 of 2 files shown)


image_provisioner/playbooks/roles/repo-install/templates/etc/yum.repos.d/all.repo.j2
```diff
@@ -53,15 +53,6 @@ enabled=1
 gpgcheck=0
 sslverify=0
 
-# This htb repo gets us openvswitch-2.6, and is required until the htb bits are promoted to the above GA channel (GA is openvswitch-2.5 for now).
```
wabouhamad: do we need to remove this repo, in case we need other packages from it in future ?

jeremyeder:
@mffiedler better file a separate issue in the svt repo with your thoughts/plan, and then start an on-list discussion about version branches of this repo...

mffiedler:
Branch pre-3.10 can be used to provision images requiring OVS

PR #434, opened 2018-04-20
Modify conformance to work for ocp 3.10

chaitanyaenr:
This commit modifies the script to restart the api and controller
pods instead of services.


mffiedler:
Looks good.  I also saw an issue with python:latest imagestream not available.  We might have to work around that.  I will try to recreate on a new install and file a bug if I see it again.

sjug:
Did it have to be a `shell` command?

jeremyeder:
yes because of these shims ...

PR #436, opened 2018-04-23
Modify openshift_labeler to use roles defined in openshift cluster

chaitanyaenr:
Instead of relying on the inventory to determine the role the node
plays in openshift cluster, use the labels defined. This commit also
modifies it to work on scale-ci env.


chaitanyaenr:
@wabouhamad @mffiedler PTAL.

PR #437, opened 2018-04-23
Update imagestream workaround for jenkins for 3.10

mffiedler:
In 3.10, the jenkins "latest" tag in the imagestream is an alias for the v3.10 image.  However this image is not yet available on registry.access.redhat.com.   As we did in 3.9, update the imagestream to have "latest" point to the actual "latest" tag in the registry.

This allows the build-related conformance tests to pass.

See:  https://bugzilla.redhat.com/show_bug.cgi?id=1535099 and internal mailing lists for discussion.


mffiedler:
@wabouhamad Can you do a quick review?

PR #438, opened 2018-04-24
Change the stepsize and pause params for nodevertical test

chaitanyaenr:
This is needed to bring down the test time to around 20 min. With
the current config, most of the time is spent in step delays.


chaitanyaenr:
@wabouhamad @mffiedler PTAL.
/cc @jeremyeder

chaitanyaenr:
@wabouhamad @mffiedler makes sense to me, we can run some tests to see if the config changes are okay.

mffiedler:
Merging since this is just tuning.  We can revisit as necessary.

PR #442, opened 2018-04-27
Remove deprecated oc get option

mffiedler:
Remove deprecated flag from oc get command


mffiedler:
Tested against 3.10.0-0.28.0 - test runs correctly again

PR #444, opened 2018-05-01
Add test_branch to pbwedge

sjug:
To separate the different test types we will hack git and use branches.


sjug:
Added force as an insurance policy

sjug:
@openshift/svt please review XS change

PR #446, opened 2018-05-02
Bake in more packages

chaitanyaenr:
Jiri found out the list of packages that are missing from the gold
image that openshift-ansible installs during the ocp installation.
This commit takes care of installing those.


chaitanyaenr:
@jmencak @wabouhamad PTAL.

wabouhamad:
LGTM

PR #452, opened 2018-05-15
Add ifexists field to master vert configs

sjug:
Master vert configs lack new mandatory field.

### DIFF
--- a/openshift_scalability/config/golang/pyconfigMasterVertScale.yaml
+++ b/openshift_scalability/config/golang/pyconfigMasterVertScale.yaml
@@ -4,6 +4,7 @@ ClusterLoader:
   projects:
     - num: 5
       basename: clusterproject
+      ifexists: reuse
       templates:
         -
           num: 3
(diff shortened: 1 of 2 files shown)


openshift_scalability/config/golang/pyconfigMasterVertScale.yaml
```diff
@@ -4,6 +4,7 @@ ClusterLoader:
   projects:
     - num: 5
       basename: clusterproject
+      ifexists: delete
```
chaitanyaenr: For master vertical scale test, we don't delete the projects until the end of all iterations. Instead of deleting the projects created in the first iteration, we add more projects to match the total count in order to avoid duplication and  long test duration.

mffiedler:
Closing based on @chaitanyaenr comment.

sjug:
@mffiedler It's a mandatory field, so while `delete` isn't the right parameter we still need to change the conifg in order for it to work.

mffiedler:
Reopening

sjug:
Changed to `ifexists: reuse`

PR #455, opened 2018-05-22
Add branches to pbwedge

sjug:
We will use branches to differentiate between test type.


sjug:
Right this was done already, https://github.com/openshift/svt/pull/444

PR #457, opened 2018-05-29
Add support for nodevertical to schedule pods on labeled nodes

chaitanyaenr:
- This will schedule the pods on nodes with a nodevertical=true label.
  This is needed to hit the max pod count on a node when the cluster
  size is greater than 4.
- This change supports labeling nodes both on Alderaan and AWS
  environments.

### DIFF
--- a/openshift_scalability/config/golang/nodeVertical-labeled-nodes.yaml
+++ b/openshift_scalability/config/golang/nodeVertical-labeled-nodes.yaml
@@ -1,0 +1,21 @@
+provider: local
+ClusterLoader:
+  cleanup: true
+  projects:
+    - num: 1
+      basename: clusterproject
+      tuning: default
+      ifexists: delete
+      pods:
+        - num: 1000
+          image: gcr.io/google_containers/pause-amd64:3.0
+          basename: pausepods
+          file: pod-pause-labeled-nodes.json
+  tuningsets:
+    - name: default
+      pods:
+        stepping:
+          stepsize: 50
+          pause: 60
+        ratelimit:
+          delay: 0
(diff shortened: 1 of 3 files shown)


chaitanyaenr:
@wabouhamad PTAL.

openshift_scalability/config/golang/nodeVertical-labeled-nodes.yaml
```diff
@@ -0,0 +1,21 @@
+provider: local
+ClusterLoader:
+  cleanup: true
```
wabouhamad: can we set cleanup to false ? So we can keep the pods deployed for 10 minutes before the project clusterproject is terminated.

openshift_scalability/config/golang/nodeVertical-labeled-nodes.yaml
```diff
@@ -0,0 +1,21 @@
...
+  projects:
+    - num: 1
+      basename: clusterproject
+      tuning: default
+      ifexists: delete
+      pods:
+        - num: 1000
```
wabouhamad: Do we want to keep default number of pods at 500 aka 250 pods per node ?

chaitanyaenr: We pick 4 app nodes, so assuming 250 pods/node, we need to create 1000 pods in total.

openshift_scalability/config/golang/nodeVertical-labeled-nodes.yaml
chaitanyaenr: I think introducing a sleep param in clusterloader which when defined should sleep before deleting the namespace is the way to go. What do you think?

mffiedler:
I don't see a reason to not have separate configs appropriate for different environments.   Can we make a separate config for the SVT pipeline?   Not advocating proliferation of configs, but where it makes sense.

wabouhamad:
@mffiedler  we can have a separate config for SVT pipeline in a separate PR.

chaitanyaenr:
@mffiedler @wabouhamad is the PR ready for merge?

PR #461, opened 2018-06-07
Bake in github token needed to upload scale-ci results

chaitanyaenr:


mffiedler:
@wabouhamad Please review

image_provisioner/playbooks/roles/os-kickstart/tasks/main.yaml
```diff
@@ -141,6 +141,12 @@
     - { find: "^#?MaxStartups.*", replace: "MaxStartups 200:30:400" }
     - { find: "^#?UseDNS.*", replace: "UseDNS no" }
 
+- name: add token needed for uploading r2r results
+  lineinfile:
+    path: "/root/svt/utils/pbwedge/hosts"
+    regexp: "github_token"
+    line: "github_token={{ scale_ci_results_token }}"
+
```
wabouhamad: Can we add the condition "when: scale_ci_results_token is defined" so we don't have to update all our AWS inventories files if we are not running scale-ci jobs ?

chaitanyaenr: Done.

PR #462, opened 2018-06-07
Add kraken tool to svt

chaitanyaenr:
Kraken is a resiliency tool for OpenShift which can crash a node,
master, etcd and checks if the cluster recovers from the failures.
This is not supported on openshift > 3.9 as the master and etcd as
this tool doesn't take into account that the master services and etcd
are static pods from ocp 3.10.


chaitanyaenr:
@wabouhamad PTAL.

wabouhamad:
@chaitanyaenr  lets get these initial files into SVT repo, LGTM.  I will create a follow up PR for changes I made to run on 3.10.  Do you need to also commit the OpenShift-Labeler code ?

PR #468, opened 2018-06-12
Added pre-3.10 kraken source code, README, and config files

wabouhamad:
Kraken is a resiliency tool for OpenShift which can crash a node,
master, etcd and checks if the cluster recovers from the failures.
This is the tool version that is supported for pre-3.10 OCP version.
- created kraken directory with subdirectories, source files, config files and png image files


wabouhamad:
@chaitanyaenr PTAL, this is the original code with updated Openshift-Labeler label.yaml playbook.

openshift_tooling/kraken/setup-env.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+if [[ $(echo $GOPATH) == "" ]];then
+	mkdir -p $gopath_dir/.go
+	echo "GOPATH=$gopath_dir/.go" >> ~/.bashrc
+	echo "export GOPATH" >> ~/.bashrc
+	echo "PATH=\$PATH:\$GOPATH/bin # Add GOPATH/bin to PATH for scripting" >> ~/.bashrc
+	source ~/.bashrc
+fi
+# get kube-monkey
+go get github.com/asobti/kube-monkey
```
chaitanyaenr: @wabouhamad we don't need this script as we are not using kube-monkey in our tool.

wabouhamad: @chaitanyaenr, removed setup-env.sh.  Should be ready for merge.

chaitanyaenr:
Thanks for the PR @wabouhamad.

PR #471, opened 2018-06-14
add simple stress tests

mrsiano:


mrsiano:
@openshift/scale-performance-review @mffiedler PTAL

openshift_scalability/content/pod-stress-simple.json
```diff
@@ -0,0 +1,37 @@
...
+      {
+        "name": "stress",
+        "image": "progrium/stress",
+        "ports": [
+          {
+            "containerPort": 8080,
+            "protocol": "TCP"
+          }
+        ],
+        "command": ["stress", "--cpu", "2", "--io", "1", "--vm", "2", "--vm-bytes", "128M"],
```
jeremyeder: As-is, this will put 2 full cores of load, that's not going to work in our infra.  There are ~2000 cores total.  The purpose of stress is to put "some data" into prometheus around cpu and memory, not to peg all cores.

So I would suggest using a different version of this.  The 'stress-ng' package (from EPEL) allows for fractional CPU load to be put on. For example, this will put 5% CPU load per pod.

```stress-ng --cpu 1 --cpu-load 5```

Comment (mffiedler) [reply openshift_scalability/content/pod-stress-simple.json]: See also:  https://github.com/openshift/svt/tree/master/openshift_scalability/content/centos-stress

That container did become a bit of a dumping ground for stress tools (jmeter, mb, stress).  Maybe create a new first-class stress-ng pod for cpu/memory/oom/etc testing.

**Comment by mrsiano** [comment]
@mffiedler can merge ?

File: openshift_scalability/content/pod-stress-simple.json [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+      {
+        "name": "stress",
+        "image": "lorel/docker-stress-ng",
+        "ports": [
+          {
+            "containerPort": 8080,
+            "protocol": "TCP"
+          }
+        ],
+	"command": ["stress-ng", "--cpu", "1", "--cpu-load", "5", "--vm", "1"],
```
mffiedler: Can we please make the command a parameter and have the config file set the default command.   See https://github.com/openshift/svt/blob/master/openshift_scalability/content/logtest/logtest-rc.json#L81 and https://github.com/openshift/svt/blob/master/openshift_scalability/config/ocp-logtest.yaml#L11 for an example

mffiedler: Ok, we can leave it as a hard-coded simple pod.   Followup might be to make it a template for a deployment or replication controller in the future.

PR #472, opened 2018-06-14
Add podvertical scale test

chaitanyaenr:
This commits adds a new scale test called podvertical which creates
maximum number of pods possible per namespace across the cluster.

### DIFF
--- a/openshift_scalability/config/cluster-limits-pods-per-namespace.yaml
+++ b/openshift_scalability/config/cluster-limits-pods-per-namespace.yaml
@@ -1,0 +1,26 @@
+projects:
+  - num: 1
+    basename: clusterproject
+    ifexists: delete
+    tuning: default
+    pods:
+      - total: 3000
+      - num: 100
+        image: gcr.io/google_containers/pause-amd64:3.0
+        basename: pausepods
+        file: content/pod-pause.json
+        storage:
+          - type: none
+          
+quotas:
+  - name: default
+    file: default
+
+tuningsets:
+  - name: default
+    pods:
+      stepping:
+        stepsize: 150
+        pause: 60 s
+      rate_limit:
+        delay: 0 ms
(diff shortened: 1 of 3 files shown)


chaitanyaenr:
@wabouhamad PTAL.
/cc @jeremyeder

jeremyeder:
This should take roughly a half hour, right?  If so, that sounds good.

Also can you please rename it something more obvious, like

cluster-limits-pods-per-namespace.yaml

openshift_scalability/config/golang/podVertical.yaml
```diff
@@ -0,0 +1,21 @@
...
+      pods:
+        - num: 3000
+          image: gcr.io/google_containers/pause-amd64:3.0
+          basename: pausepods
+          file: pod-pause.json
+  tuningsets:
+    - name: default
+      pods:
+        stepping:
+          stepsize: 150
```
wabouhamad: The stepsize is increased 3 times, compared to 50 in node vertical, I am assuming it is to keep overall test duration short.  Have we tested if we can repeatably sustain deploying 150 pods with 60 sec pause time, I am assuming this will be spread across multiple compute nodes.

jeremyeder: great question.  and also from a procedural standpoint it seems like we merge tests before they are themselves tests.  I don't have a huge objection to this (or suggested fix), but it does raise valid questions like @wabouhamad raised.  For now, if it does cause problems, we follow up with another PR tweaking it to the values that ended up working?

chaitanyaenr: We have tripled the step size to keep the test suration short. I have tested this on a 5 node aws cluster and the cluster loader was okay with creating 150 pods at a time and it went on till we ran out of resources on the cluster to create more pods.

PR #474, opened 2018-06-15
pbwedge pretty print json before commit

sjug:
When creating new branches, get rid of extra echo for README text.
Also jq the output so that the result is also somewhat human readable at a glance.


sjug:
@mffiedler Can you give this a quick look :smiley:

PR #475, opened 2018-06-19
Introduce a prometheus loader tool.

mrsiano:
the following tool allow us to generate some workloads against promethues specially for openhift stack.

motivation:
come up with a tool that allow us to simulate the grafana workloads.
the tool fire prometheus queries simultaneously over http against the prometheus server as the grafana dose

### DIFF
--- a/openshift_scalability/PrometheusLoader.md
+++ b/openshift_scalability/PrometheusLoader.md
@@ -1,0 +1,14 @@
+# Intro
+This tool is used to fire prometheus queries against a prometheus instance in OpenShift.
+
+It also helps to simulate Grafana workloads on a staging cluster.
+
+The app uses a query file which includes a collection of prometheus queries,
+the collection fits only to OpenShift deployment.
+
+# How to
+python prometheus-loader.py -f <query_file> -i <interval> -t <concurrency> -p <graph_period>  > /dev/null 2>&1 &
+
+interval = bulk frequency, how often queries were being fired (in seconds).
+concurrency = how many queries will work in parallel (0-100).
+period = for what time period \ data size (in minuets).
--- a/openshift_scalability/ci/scripts/prometheus-loader.sh
+++ b/openshift_scalability/ci/scripts/prometheus-loader.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+set -x
+
+refresh_interval=$1
+concurrency=$2
+graph_period=$3
+duration=$4
+enable_pbench=$5
+pbench_copy_results=$6
+pbench_user_benchmark=$7
+
+cd /root/svt/openshift_scalability
+# TODO: get the script file, remove this once the PR merged.
+wget https://raw.githubusercontent.com/mrsiano/svt/prometheus_loader/openshift_scalability/prometheus-loader.py
+wget https://raw.githubusercontent.com/mrsiano/svt/prometheus_loader/openshift_scalability/content/prometheus/qs.txt
+
+oc login -u system:admin
+if [ "${enable_pbench}" == "true" ]; then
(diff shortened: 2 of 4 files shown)


mrsiano:
@jeremyeder @mffiedler

mrsiano:
I've tested this tool and I'm already using it, I really can see it affect the prometheus performance.

jeremyeder:
can you throw together a README about how to use it?
is @brancz ok with the approach?

brancz:
Generally this approach seems fine. Just a few minor questions that could also be follow ups:

* How did you choose the queries? Ideally I’d like to use the queries that the Grafana dashboards generate (we merged a Grafana integration in the cluster monitoring stack a few days ago).
* While some constant load seems fine we shouldn’t overdo it with the concurrency when we do the actual tests. I didn’t check the queries super thoroughly but it’s for sure possible to kill a prometheus Server if you do a brutally complex query - like any other database. We should make sure it’s a good mix of being realistic and relatively constant. Remember that we already have constant query load because of recording rules, what we’re trying to simulate here is ad hoc (random query but rather rarely executed) and dashboard querying (typically dashboards reload data every 10s).

jeremyeder:
just out of curiosity why would our default dashboard reload interval be any smaller than the scrape interval?  (or even 2x the scrape interval?)  obviously users can dial it down to something lower.

mrsiano:
@brancz 
regarding your questions:
- those query I'v collected to the qs.txt file being collected from here x which is our grafana dashboard for several diffrent CI clusters.
- about the concurrency that how the grafana works, in case you have few graphs in your dashboard and you refresh it you will see simultaneous requests by any network tracer.
image
about the complexity it depends on what data you wanted to draw, in our case we developed some graphs that needs data aggregation and some ohter math functions (BTW, it's pretty much basic and not so complex).
any way we can tune that qs.file and add or remove that collection accordingly.

@jeremyeder 
currently the default is 20 sec, but the user can setup what ever he want and control it by the command line arg, I can hard code a minimum  value to guarantee the end use will not overkill it.

brancz:
@mrsiano got it, all of that sounds good to me, thanks for clarifying.

@jeremyeder I'm not sure I understand how querying is related to scrape interval here. Can you elaborate what you meant?

jeremyeder:
I meant from a practical standpoint as a user, that isn't it useless to have your dashboard refresh interval lower than the scrape interval?  No new data will show up?

mrsiano:
@mffiedler can you merge this please ?

openshift_scalability/PrometheusLoader.md
```diff
@@ -0,0 +1,15 @@
+# Intro
+this tools used to fire prometheus queries against prometheus instance in OpenShift
```
mffiedler: Editorial:  "This tool is used to fire prometheus queries against a prometheus instance in OpenShift."

openshift_scalability/PrometheusLoader.md
```diff
@@ -0,0 +1,15 @@
...
+stack context.
+
+It's also helps to simulate Grafana workloads, on staging cluster.
```
mffiedler: Editorial:  "It also helps  to simulate Grafana workloads on a staging cluster."

openshift_scalability/PrometheusLoader.md
```diff
@@ -0,0 +1,15 @@
...
+
+the app uses query file which include a collection of prometheus queries,
+the collection fits only to OpenShift context
+
```
mffiedler: Editorial:  "The app uses a query file which includes a collection...."

mrsiano:
@mffiedler any blocker here ?

PR #477, opened 2018-06-20
update test to push standard sized alderaan cluster to its pod density limit

jeremyeder:
Update this test to push cluster roughly to it's pod density limit.

5,000 pods per ns pushes beyond the 3,000 limit we have for 3.10 so we have some headroom.


jeremyeder:
@chaitanyaenr

jeremyeder:
I'm going to close this until we think through CI of the whole limits table.

PR #478, opened 2018-06-21
Add support to install crio rpm and needed tools

chaitanyaenr:
This commit modifies the cri-o role to support installing crio as a
system container and rpm. This also moves the r2r token bake in step 
to pbench-kickstart as it needs the svt repo to be cloned to modify 
the token.


chaitanyaenr:
@wabouhamad makes sense, updated the PR to set the mount point to /var/lib/containers by default when crio=True.

mffiedler:
@wabouhamad good to merge?

PR #479, opened 2018-06-25
Update node labeling on svt openshift_scalability CI scripts for 3.10 and pod limits to stay under 250 per node

wabouhamad:
- Updated node labeling on svt openshift_scalability CI scripts for 3.10 to use node-role.kubernetes.io label on various scripts used in CI SVT Pipeline
- updated pod counts on golang cluster loader config files to stay under the default kubeletArgument max-pods 250 pods per node value
- Updated number of pods from 250 to 240 for:
   - node-affinity-anti-affinity
   - pods-affinity-anti-affinity
   - taints and tolerations
- Created a new NodeVertical yaml config file with 484 pause pods to help total stay under 250 pods per node
- Updated Node Vertical script to set cleanup to False in nodeVertical.yaml script during test and then revert it back to original after the test

### DIFF
--- a/openshift_scalability/ci/scripts/pbench-controller-start-node-vertical.sh
+++ b/openshift_scalability/ci/scripts/pbench-controller-start-node-vertical.sh
@@ -81,12 +81,10 @@ pbench-clear-results
 
 
 # get the internal node ip addresses now that we have the /root/.kube/config copied over:
-## NODES=$(oc get nodes | grep -v "SchedulingDisabled" | grep -v NAME | awk '{print $1}'  | xargs)
 ## NODES are used when we want to register pbench from scratch
 NODES=$(oc get nodes | grep -v "master" | grep -v NAME | awk '{print $1}'  | xargs)
 
 ALL_NODES_INTERNAL=$(oc get nodes | grep -v NAME | awk '{print $1}'  | xargs)
-## MASTER_INTERNAL_IP=$(oc get nodes | grep "SchedulingDisabled" | awk '{print $1}')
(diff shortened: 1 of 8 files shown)


wabouhamad:
@mffiedler @chaitanyaenr  PTAL

jeremyeder:
why can't we run 250?  is it because of all the control plane / infra pods we run now? (sdn/node_exporter...) ?

chaitanyaenr:
@jeremyeder you are right, there will be some pods running on the node including prometheus-node-exporter and openshift sdn. So, we need to set the pod count taking into account the number of pods which are already running on the node.

@wabouhamad @jeremyeder I think it's better to add a small logic to calculate the number of pods on the labeled nodes where the nodevertical schedules the pods and then replace the value in the config. This way we need not manually look at the cluster to set the pod count in the config and also make sure we hit the max pod limit i.e 250.

wabouhamad:
@jeremyeder yes, prometheus-0, node exporter, sync, sdn, ovs pods are already running on compute nodes.  KubeletArgument max-pods is set to 250 by default and changing it requires editing the compute node configmap or creating custom openshift-node-groups with new values set in the node group Edits field at install time.

wabouhamad:
@jeremyeder @chaitanyaenr  Additionally for node affinity/anti-affinity and taints/toleration tests, the pods I am running have memory and cpu requests, and at 250 pods we were reaching node cpu capacity.

jeremyeder:
Ack - then perhaps our cluster_limits doc needs to have a note added to it that you can run 250 pods max, minus ocp components, which yields what - 245 "user" pods per node?

wabouhamad:
@jeremyeder I agree, the note about the ~ 245  "user" pods per node in the cluster_limits doc would be useful.  I can create an issue for that.

PR #481, opened 2018-06-26
Modify the max pods per ns test to create 15000 pods

chaitanyaenr:
This commit also increases the step size to 500 and decreases the pause time
to 30 seconds so as to keep the test run duration under 90 min.


chaitanyaenr:
@jeremyeder PTAL.

jeremyeder:
@tkatarki

jeremyeder:
lgtm

PR #482, opened 2018-06-29
Modify max pods per ns test to support multiple iterations

chaitanyaenr:
This commit modifies the scripts and config's to not delete the project created so that we can continue to create more pods on top of the existing pods.


chaitanyaenr:
@mffiedler PTAL.

PR #495, opened 2018-07-23
Updates to start network test

schituku:
As requested by @chaitanyaenr updated the following.
  -- removed the config file parameter from the jenkins job
  -- removed the pub key and copying it from known location


schituku:
more commits made into the branch than necessary. Hence closing it.

PR #497, opened 2018-07-24
Conformance wrapper updates for 3.11

mffiedler:
1. Make the Jenkins image version substitution version independent, at least as long as this approach for the jenkins v2 images is in place.
2. Also, enable atomic-openshift-excluder so that tests can be installed, if not present.   Disable when done.
3. 30 second sleep after restarting master - investigate difference with 3.10 as a FUP
4. Skip idling test - it has been failing in AfterSuite teardown since 3.10.  Needs investigation.


mffiedler:
@chaitanyaenr PTAL

PR #498, opened 2018-07-24
Fix for multi master environment

schituku:
-- earlier version assumed that there will be only one master, this
fix will work for both single and multi-master environment.


jeremyeder:
Thank you for the quick turnaround, Siva!

PR #506, opened 2018-08-01
Use the default config if not defined by the user in podvertical

chaitanyaenr:


openshift_scalability/podVertical.sh
```diff
@@ -12,6 +12,14 @@ fi
 TYPE=$1
 CONFIG=$2
 
+# use the default config if it's not defined by the user
```
mrsiano: may be it worth  set the golang as default if both python or golang will not mentioned ?
currently it will break if it's not specified.

something like:
if -z config:
case $type in
golang)
  ;;
python)
  ;;
*)
  # type golang
  # config golang
  ;;
esac
fi

PR #510, opened 2018-08-13
Increase vm.max_map_count for elasticsearch 5

mffiedler:
@wabouhamad @chaitanyaenr PTAL and merge with any private branches being used currently for provisioning.

This is for Elaticsearch 5.   According to https://access.redhat.com/solutions/99913 this should not have a negative impact for non-ES nodes - it just allows a larger footprint.

/cc: @jeremyeder

### DIFF
--- a/image_provisioner/playbooks/roles/os-kickstart/tasks/main.yaml
+++ b/image_provisioner/playbooks/roles/os-kickstart/tasks/main.yaml
@@ -149,3 +149,12 @@
 
 - name: install pytimeparse module
   command: pip install pytimeparse
+
+- name: update vm.max_map_count
+  lineinfile: 
+    dest: /etc/sysctl.conf
+    line: "vm.max_map_count=262144"
+    state: present
+    insertafter: EOF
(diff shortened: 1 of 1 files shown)


mffiedler:
Updated to remove test directory and point to /etc/sysctl.conf.

jeremyeder:
isnt this already done by openshift-ansible for 3.11? @ewolinetz

ewolinetz:
yes, this is done already in the openshift-logging playbook [1] for 3.11 if the value is less than the required amount of if it is omitted

[1] https://github.com/openshift/openshift-ansible/blob/master/playbooks/openshift-logging/private/config.yml#L95-L113

jeremyeder:
ok. if the product does it, then we should not need this pr

mffiedler:
Installed 3.11.0-0.13.0 today and it did not get set.  Will investigate further.

mffiedler:
Closing this PR - verified the logging deployer takes care of this in the latest puddles.

PR #511, opened 2018-08-16
Add support to install pbench-agent-internal

chaitanyaenr:
This commit installs pbench-agent-internal to get the keys and config
instead of modifying the example config.


chaitanyaenr:
Thanks for the review @wabouhamad @mffiedler.

PR #513, opened 2018-08-23
Modify nodevertical cleanup to wait for resources to be deleted

chaitanyaenr:
This commit adds a wait flag to the nodevertical scale test to make
sure the resources created by the test are cleaned up.


chaitanyaenr:
@wabouhamad @mffiedler PTAL.

PR #520, opened 2018-09-12
Add --raw option with no prefix

mffiedler:
Add a --raw option that does not prepend anything to the message.

@schituku please review

### DIFF
--- a/openshift_scalability/content/logtest/ocp_logtest-README.md
+++ b/openshift_scalability/content/logtest/ocp_logtest-README.md
@@ -40,7 +40,7 @@ login to the host where the pod is running and verify the logs are going to jour
 
 ```python ocp_logtest.py <optional-arguments>```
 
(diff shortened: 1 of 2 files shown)


pruan-rht:
LGTM

mffiedler:
@anpingli found a problem with the line lengths when reading from files.  Need to update the PR to make raw mode use file lines as-is regardless of line length settings.  I pushed a container image that has a temporary solution but need a final solution that is not so hacky.

mffiedler:
Updated commit  to only return the line read when the mode is raw.   Don't try to adjust the line length.

mffiedler:
@schituku found a few issues

1. README needs to be  updated
2. The --raw flag still has some strange interactions with other options.   Need to make --raw exclusively work with --file and ignore other options like fixed and line length

mffiedler:
@schituku PR updated.  Please take a look.

anpingli:
Great

PR #524, opened 2018-09-19
Updated the OpenShift quickstart dancer-mysql templates to the latest version from OCP 3.11.3-1

wabouhamad:
Dancer-mysql builds were failing when building DBD-mysql.  Templates were updated from the latest OCP 3.11-3 templates in the openshift namespace.


wabouhamad:
@mffiedler PTAL

PR #525, opened 2018-09-19
Updated build_ami.yaml to append the openshift_version string to ami builder instance name

wabouhamad:
We have had automated jobs trigger from UMB messages for different OpenShift puddles and using the same ami builder instance.  This PR will make the ami builder instance name more unique to each puddle, by appending the openshift_version string to the instance name.


wabouhamad:
@mffiedler, @chaitanyaenr PTAL

PR #528, opened 2018-09-25
Improvements to intermediate mode

schituku:
-- improve int mode to run faster
  -- update docker to install latest epel
  -- remove clearing pbench tools so it can run smoothly in scale lab


schituku:
@chaitanyaenr PTAL

PR #529, opened 2018-10-02
Automation of router tests

schituku:
-- run the router test
-- analyze the results
-- copy them to pbench server
-- generate graphs for the results


mffiedler:
Verified tests run successfully and graphs are generated.   Data looks reasonable to me.

PR #531, opened 2018-10-08
ignore sleep.

mrsiano:
ignore long sleep on demand for pod vertical.

the motivation comes from some prometheus workloads that dose not need steady state

### DIFF
--- a/openshift_scalability/podVertical.sh
+++ b/openshift_scalability/podVertical.sh
@@ -11,19 +11,24 @@ fi
 
 TYPE=$1
 CONFIG=$2
+IGNORE_LONG_SLEEP=$3
 
 # use the default config if it's not defined by the user
 if [[ -z "$CONFIG" ]] && [[ "$TYPE" == golang ]]; then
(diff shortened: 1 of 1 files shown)


mrsiano:
@chaitanyaenr @mffiedler

openshift_scalability/podVertical.sh
```diff
@@ -11,19 +11,24 @@ fi
 
 TYPE=$1
 CONFIG=$2
+IGNORE_LONG_SLEEP=$3
 
 # use the default config if it's not defined by the user
 if [[ -z "$CONFIG" ]] && [[ "$TYPE" == golang ]]; then
 	CONFIG=config/golang/cluster-limits-pods-per-namespace
```
chaitanyaenr: Can you also fix the spacing here just to be consistent?

jeremyeder:
This only saves 6 minutes per test - do we really need a special case?

mffiedler:
@mrsiano Going to close this based on the comments unless you feel strongly about pushing it forward.

PR #534, opened 2018-10-25
Add a check to make sure the nodes are labeled

chaitanyaenr:
This commit:
- Modifies the script to fail in case there are zero nodes
labeled with pbench_role=agent or zero pods running in the pbench
namespace.
- Avoids tagging the pbench images for the templates to work out
of the box.

### DIFF
--- a/openshift_templates/performance_monitoring/pbench/pbench-agent-daemonset.yml
+++ b/openshift_templates/performance_monitoring/pbench/pbench-agent-daemonset.yml
@@ -15,9 +15,9 @@ spec:
       hostPID: true
       hostNetwork: true
       containers:
-      - image: library/pbench-agent:latest
+      - image: docker.io/ravielluri/image:agent
         name: pbench-agent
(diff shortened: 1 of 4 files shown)


chaitanyaenr:
@mffiedler PTAL.

openshift_templates/performance_monitoring/pbench/pbench-agent-daemonset.yml
```diff
@@ -15,9 +15,9 @@ spec:
       hostPID: true
       hostNetwork: true
       containers:
-      - image: library/pbench-agent:latest
+      - image: docker.io/ravielluri/image:agent
```
mffiedler: We should move test images to quay.io.  Not a blocker for this PR.   Will open an issue.

chaitanyaenr: Sounds good, will move the image to quay.io. Thanks for the review @mffiedler.

PR #537, opened 2018-11-09
Updated aos-ansible hosts template to set correct registry folder name when pre-pulling images from registry.reg-aws to support ocp v4 and remain backwards compatible with OCP v3

wabouhamad:
Ose images for OCP v3 are in the "openshift3 folder on registry.reg-aws.openshift.com
while images for OCP v4 are in the "openshift" folder.  This PR sets the correct folder name based on the puddle version in the aos-ansible hosts file template hosts.j2.


wabouhamad:
@chaitanyaenr @mffiedler PTAL

PR #540, opened 2018-11-21
Allow skipping of registering pbench tools.

akrzos:
Running the network tests in scale-ci causes tools to be re-registered and in
specific occasions, it can pollute the names of pbench results. This enables
avoiding that situation with the network tests.


mffiedler:
@schituku please take a look

akrzos:
@schituku PTAL and double check if I made the requested changes.

PR #541, opened 2018-11-28
Always use labeled nodes for pause pods with NV

akrzos:
With OSA testing, I have found pods on nodes that were not labeled, thus pods
were spread across more nodes than expected.


akrzos:
@chaitanyaenr  Please take a look

akrzos:
@mffiedler Could we get this merged?

PR #554, opened 2019-02-12
Adding jump_to_node script

skordas:
Simple script to jump to internal worker node


mffiedler:
@skordas I believe this is obsolete now.  OK for me to close it?

skordas:
@mffiedler Yes, not used anymore. 
`oc debug node` is good solution.

PR #556, opened 2019-02-21
Minor fixes in networking test

chaitanyaenr:
This commit modifies the networking test to move results instead of
copy and groups all the results under networking prefix.


chaitanyaenr:
@akrzos PTAL.

akrzos:
LGTM

PR #561, opened 2019-03-12
Use exit code from ansible playbook run

smalleni:
Earlier the exit code was only being printed. This commit enhances
the code to use the exit code from the ansible playbook run and notify
user in case of errors. In cases where several playbooks are run, each for
a different pod number if one playbook fails, the earlier playbooks also
tend to fail due to uncleand projects from earlier runs. In the future
if needed we can exit from the loop running playbooks based on the exit code
of the previous playbook.


smalleni:
@chaitanyaenr @vikaslaad  PTAL

smalleni:
@chaitanyaenr Any update?

PR #566, opened 2019-03-18
monogdb test updated 

ekuric:
Changes

- With dc.yaml it was noticed that ycsb pods will not be equally spread across nodes, this is experiment for now. Test works with this too, but when pod object is used to create pods , pods are spread more equally than for case when dc.yaml is used. 
-  updated files/scripts/create-oc-objects.sh to delete project before new ones are created
I noticed in tests that having projects deleted in advance before creating new ones ( for new test ) gives healthier environment  ( especially for case when storage backend is not good with cleaning stuff properly ) when projects are deleted first and then new test start from clean state - without any projects 
-  updated runmongo.sh to support new options

### DIFF
--- a/storage/mongodb/files/oc/ycsb_pod.yaml
+++ b/storage/mongodb/files/oc/ycsb_pod.yaml
@@ -1,0 +1,21 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    name: ycsb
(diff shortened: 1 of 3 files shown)


ekuric:
@hongkailiu @mffiedler ptal and comment, if all fine, please merge

PR #569, opened 2019-03-28
Tooling enchancements

chaitanyaenr:
This commit:
- adds support to run and collect pbench data for conformance and scaleup
  using containerized tooling.
- mounts / into the pbench-agent and controller pods to be able to collect
  info about the pvc's, mainly the ones by prometheus.

### DIFF
--- a/dockerfiles/pbench/controller/run.sh
+++ b/dockerfiles/pbench/controller/run.sh
@@ -63,8 +63,19 @@ elif [[ $JOB == "http" ]]; then
 	git clone https://github.com/jmencak/http-ci-tests.git /root/http-ci-tests
 	cd /root/http-ci-tests
 	source /opt/pbench-agent/profile; . ./http-test.sh all
-elif [[ $JOB == "mongo" ]]; then
-	source /opt/pbench-agent/profile; 
+elif [[ $JOB == "scaleup" ]]; then
+        echo "Running ocp scaleup"
+        git clone https://github.com/redhat-performance/scale-ci-ansible.git /root/scale-ci-ansible
+        cd /root/scale-ci-ansible
+	cat <<-EOF >> scaleup.inv
+	[orchestration]
+	localhost ansible_connection=local
+	EOF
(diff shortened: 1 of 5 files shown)


chaitanyaenr:
@akrzos  @mffiedler PTAL.

dockerfiles/pbench/controller/run.sh
```diff
@@ -63,8 +63,19 @@ elif [[ $JOB == "http" ]]; then
 	git clone https://github.com/jmencak/http-ci-tests.git /root/http-ci-tests
 	cd /root/http-ci-tests
 	source /opt/pbench-agent/profile; . ./http-test.sh all
-elif [[ $JOB == "mongo" ]]; then
-	source /opt/pbench-agent/profile; 
+elif [[ $JOB == "scaleup" ]]; then
+        echo "Running ocp scaleup"
+        git clone https://github.com/chaitanyaenr/scale-ci-ansible.git /root/scale-ci-ansible
```
akrzos: In practice we should be tying this to the redhat-performance namespace rather than individual user namespace.  Is there a specific requirement at this time to clone from local fork rather that the original repo?  To extend this, I would like to get away from the practice of git cloning in code as it leads to non-deterministic/greater difficulty in reproducing any tests.  I understand from a realistic sense right now we can not necessarily get away from that practice as it is tech debt at this point.

chaitanyaenr: scale-ci-ansible has all the stuff we need for scaleiup, I didn't notice that it was using my namespace. Updated the PR. Thanks.

chaitanyaenr: Agree that we need to avoid using the forks wherever possible.

akrzos: Got it, Yup at least for now we can avoid forks, and work on reducing (if possible the git cloning inside of tests in the future.)

chaitanyaenr:
Thanks for the review @akrzos.

PR #573, opened 2019-04-05
Remove the root mount from controller, pbench-agent templates

chaitanyaenr:
The reason for having the / volume mount was to be able to get info
about the pvc's but it will replace the existing contents of the
pod which is not the intended behavior.We can achieve the same
by mounting /var/lib/kubelet/pods.


chaitanyaenr:
@akrzos PTAL.

akrzos:
I tested the change earlier in the week as well and it ran great.

PR #580, opened 2019-04-23
Set the controller and pbench image policy to Always

chaitanyaenr:
This will help in creating the pods using the latest image instead
of the old image in case we have some fixes in repos baked into the
image instead of manually pulling down the image on each node.


chaitanyaenr:
@akrzos PTAL.

PR #583, opened 2019-04-26
Checking if pods are correct labeled.

skordas:

PR #584, opened 2019-05-08
Execute command on node with root user

skordas:


mffiedler:
Test with #583 
/lgtm

PR #588, opened 2019-05-16
Updated django-postgresql.json quickstart template for OCP 4.1

wabouhamad:
Updated the json template for quickstart django-postgresql app.


wabouhamad:
@mffiedler PTAL

mffiedler:
@wabouhamad I assume the other django templates should be updated as well.  True?   django-postgresql-pv.json, django-build.json, etc?

wabouhamad:
@mffiedler PTAL, I added the pv and build templates for django quickstart, and also for the new eap71 and jws31-tomcat8 apps.

PR #589, opened 2019-06-03
Adding script and templates for drain node - stress test

skordas:


skordas:
Adding script bash and config files to be able to test detach/attach PVC devices from/to compute nodes.

PR #593, opened 2019-06-06
Script to watch for pod state changes

mffiedler:
A script to monitor all running pods in the cluster and watch for changes in:

- pod start/stop
- state
- readiness
- ip
- node 
- age
- restarts

perl pod_spy.py <interval in seconds>

### DIFF
--- a/utils/scripts/pod_spy.py
+++ b/utils/scripts/pod_spy.py
@@ -1,0 +1,101 @@
+import subprocess
+import re
+import time
+import sys
+import logging
+
+# regex to parse oc get pods -o wide lines
+# cakephp-mysql-example-5      cakephp-mysql-example-1-build   1/1     Running     0          54s     10.128.2.27    ip-10-0-173-169.us-west-2.compute.internal   <none>
+pod_entry_regex = re.compile("(.*?)\s+(.*?)\s+(.*?)\s+(.*?)\s+(.*?)\s+(.*?)\s+(.*?)\s+(.*?).*")
+logger = logging.getLogger("pod_spy")
+
+
+
+def oc(cmd, config=""):
+    cmd = "oc " + cmd
+    rc = 0
+    result = ""
+    if config:
+        cmd = "KUBECONFIG=" + config + " " + cmd
+    try:
+        result = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
+    except subprocess.CalledProcessError as cpe:
+        rc = cpe.returncode
+        result = cpe.output
+    
+    string_result = result.decode("utf-8")
+    return string_result, rc
+
+def init_logger(my_logger): 
+    my_logger.setLevel(logging.INFO)
+    ch = logging.StreamHandler()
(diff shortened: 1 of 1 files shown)


mffiedler:
@skordas @RH-ematysek PTAL

utils/scripts/pod_spy.py
```diff
@@ -0,0 +1,108 @@
+import subprocess
+import re
+import time
+import sys
+import logging
+
+# regex to parse oc get pods -o wide lines
+# cakephp-mysql-example-5      cakephp-mysql-example-1-build   1/1     Running     0          54s     10.128.2.27    ip-10-0-173-169.us-west-2.compute.internal   <none>
+pod_entry_regex = re.compile("(.*?)\s+(.*?)\s+(.*?)\s+(.*?)\s+(.*?)\s+(.*?)\s+(.*?)\s+(.*?).*")
+logger = logging.getLogger("pod_spy")
```
RH-ematysek: This line isn't needed since logger is re-declared on line 98

utils/scripts/pod_spy.py
```diff
@@ -0,0 +1,108 @@
...
+            logger.info("GONE: " + pod)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        sleep_time = 15
+    else:
+        sleep_time = int(sys.argv[1])
+    logger = logging.getLogger("pod_spy")
+    init_logger(logger)
```
RH-ematysek:
```suggestion
    init_logger(logging.getLogger("pod_spy"))
```
So that 'logger' in functions don't shadow 'logger' in outer scope

utils/scripts/pod_spy.py
```diff
@@ -0,0 +1,108 @@
...
+        if not pod in new:
+            logger.info("GONE: " + pod)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        sleep_time = 15
+    else:
+        sleep_time = int(sys.argv[1])
+    logger = logging.getLogger("pod_spy")
```
RH-ematysek:
```suggestion
```

utils/scripts/pod_spy.py
```diff
@@ -0,0 +1,105 @@
...
+    except subprocess.CalledProcessError as cpe:
+        rc = cpe.returncode
+        result = cpe.output
+    
+    string_result = result.decode("utf-8")
+    return string_result, rc
+
+def init_logger(my_logger): 
+    my_logger.setLevel(logging.INFO)
+    fh = logging.FileHandler('/tmp/pod_spy.log')
```
skordas: Are we always expecting to use this file?
If yes I'm fine with hardcoded value.
But when this is related to other test where this value is not hardcoded then nicer would be to have parameter or catch exception to exit gracefully.

mffiedler: I removed the file logger - it's just a simple utility.  If someone needs a file log they can tee the output.

PR #600, opened 2019-07-10
Cleanup content templates

sjug:
In order for these templates to be consumed by the API they need to be valid. These are the templates included in the pyconfigMasterVertScalePause CL config.


mffiedler:
@sjug ++ to this.  I'm sure there are a lot more that need to be addressed - lots of alpha labels and other formatting cruft that oc process tolerates, etc but that can be done separately.   What's you opinion on converting these to yaml?   Again not for this PR, but will cluster-loader (golang) have any issues if the content is all converted to yaml?  We can trade brace/bracket madness for whitespace madness.

akrzos:
LGTM

PR #602, opened 2019-07-19
Updated playbook scale_up_complete.yaml and check_app.py to run on OCP 4.x

wabouhamad:
These updates were needed to run the scale_up_comlete.yaml playbook on OCP 4.x:
- Updated script openshift_performance/ose3_perf/scripts/check_app.py to pass 3 arguments to the login method:
- api_url
- login username
- login passwd

Updated  the openshift_performance/ci/content/scale_up_complete.yaml playbook to support passing the api_url, login_username, and login_passwd as arguments when running the playbook from ansible-playbook.  Also updated the eap scale up task to use the eap71 app, and added a new task to scale-up the jws31-tomcat8 app.

### DIFF
--- a/openshift_performance/ci/content/scale_up_complete.yaml
+++ b/openshift_performance/ci/content/scale_up_complete.yaml
@@ -1,64 +1,97 @@
 ---
-- hosts: all
+- hosts: localhost
   remote_user: root
   tasks:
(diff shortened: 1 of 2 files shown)


wabouhamad:
@mffiedler PTAL

PR #604, opened 2019-08-12
Logtest verification tool improvements

RH-ematysek:
Improvements to speed up logtest verification


RH-ematysek:
Squashed commits

PR #605, opened 2019-08-19
Add verbose option to show messages when pulling from search API

RH-ematysek:
Add an option for more verbose logging


mffiedler:
+1

PR #611, opened 2019-09-03
Added new script to run pod affinity anti-affinity testcase with new  golang cluster loader config file pod-affinity.yaml

wabouhamad:
This is a new script in:
svt/openshift_scalability/ci/scripts/run-pod-affinity-anti-affinity.sh 

to launch GoLang cluster loader from openshift-tests with the absolute path to config file:
- pod-affinity.yaml (has absolute path to the json pod template files)
- the number of pods per project was also adjusted down to 130 pods per node so we don't run out of CPU resources on m5.xlarge instances.  In 4.x,  we have system reserved CPU and several new system pods in 4.2 that have additional CPU requests

### DIFF
--- a/openshift_scalability/ci/scripts/run-pod-affinity-anti-affinity.sh
+++ b/openshift_scalability/ci/scripts/run-pod-affinity-anti-affinity.sh
@@ -1,0 +1,225 @@
+#!/bin/bash
+# set -x
+
+date
+uname -a
+openshift version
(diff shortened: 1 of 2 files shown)


wabouhamad:
@mffiedler PTAL

wabouhamad:
oops, missed you merged.  Refreshed my browser page ...

PR #615, opened 2019-10-14
Script to automate changes to machinesets

skordas:
Simple bash script that applies patches to machinesets.
Script can be used to prepare correct set of machinesets before test.


skordas:
Moved to my own repository... I think nobody will use this here.

skordas:
https://github.com/skordas/ocp-my-toolbox/blob/master/scripts/change_machineset.sh

PR #620, opened 2019-11-06
Updates for OCP 4.3: tolerate new message and allow running with just kubeconfig

mffiedler:
1.  New message when oc get pods returns no resources:

No resources found in <poroject name> namespace.

Just check that it starts with "No resources found"

2. Don't require -m/-u/-p - allow the user to set their own KUBECONFIG or ~/.kube/config


mffiedler:
@wabouhamad PTAL

mffiedler:
@wabouhamad NM - looks like I have conflicts with your recent changes.    Will re-submit

mffiedler:
OK, resolved conflict.   I went with the startswith() approach.

PR #622, opened 2019-11-20
Added --skip-config-write=true to oc new-project command in python cluster-loader utils.py

wabouhamad:
The --skip-config-write=true flag for oc new-project command will speed up bulk project creation.


wabouhamad:
@mffiedler @chaitanyaenr PTAL

mffiedler:
> LGTM, golang clusterloader skips the config write by default as well.

That's the real answer.  All golang all the time.  Still having some issues with parallel to iron out in the logging tests.

PR #623, opened 2019-11-20
Correcting verification message.

skordas:
After update https://github.com/openshift/cluster-node-tuning-operator/pull/90
there is no more information about ownerReferences, so actual message is without this part.


skordas:
@mffiedler PTAL

PR #624, opened 2019-12-10
Changing log level from DEBUG to INFO

skordas:
changing default log level for python cluster-loader.


skordas:
@mffiedler PTAL

mffiedler:
Thank you

PR #625, opened 2019-12-18
Increased loop counter for 100 to 150 waiting for pods to terminate a…

wabouhamad:
Updated two scripts:
- svt/openshift_performance/ose3_perf/scripts/scale_test.py:
   increased iteration counter from 100 to 150 to allow waiting longer for pods to terminate after scale down
- svt/openshift_performance/content/scaleup_complete.yaml:
  increased number of replicas for cakephp-mysql-example app from 20 to 100


wabouhamad:
@mffiedler PTAL

PR #626, opened 2020-01-15
Bash scripts to run HA master failover test cases

skordas:
e2e bash scripts to run HA master failover tests OCP-22423, OCP-22421


schituku:
@skordas are the tests OCP-22421, OCP-22422 ? and not "OCP-22423"?

PR #628, opened 2020-02-04
Update node tuning operator test scripts to work with OCP 4.4

skordas:
Because NTO was rewritten to work in better way I needed to update test scripts for correct verification.
There is no change how to run scripts.


skordas:
@mffiedler PTAL

mffiedler:
@skordas Should I create a 4.3 branch to preserve the previous automation.  I did not see any conditionals for release specific behavior

skordas:
@mffiedler 
yes... 3.x < current master will work here < 4.4
latest changes are working only with version >= 4.4

mffiedler:
svt-4.3 branch created

PR #629, opened 2020-02-10
Automation of TC 22140

rpattath:
This TC compares performance of oc on the version being tested with a previous version


mffiedler:
Committing this as v0.5 - we should capture the elapsed time and  compare it to some maximum from previous releases to better identify regressions, but that can be an enhancement.

PR #631, opened 2020-02-24
Storge git test update

skordas:
Update script to work with no problems with 4.x version.


skordas:
@mffiedler @rpattath PTAL

mffiedler:
@rpattath PTAL

PR #633, opened 2020-03-09
Creating script to run fio workload test with OCP 4.x

skordas:
To run: Check/edit values in external_vars file and run script:
bash run_fio.sh


mffiedler:
@skordas We should look at using ripsaw for fio workloads.   Can you investigate?   It runs fio containerized - see if it meets our requirements.    https://github.com/cloud-bulldozer/ripsaw/blob/master/docs/fio_distributed.md

PR #634, opened 2020-03-13
Super simple random openshift infra pod killer

mffiedler:
Simple, but effective, random pod killer for OpenShift infra pods.

@skordas  PTAL


skordas:
I love it!
lgtm

PR #635, opened 2020-04-01
Add templates with Deployments to replace DeploymentConfigs

mffiedler:
DeploymentConfigs are an older OCP-only resource type and require more processing for OCP to create them since they involve the creation of a deployment pod.   Deployments are the preferred application type now and can be instantiated without a deployer pod having to spin up.

1. Add deployment templates equivalent to our previous deployconfig templates
2. Update the master vert configs (python and golang) to use the new deployment templates

@chaitanyaenr I did not create a corresponding deployment-config-cluster-limits-deployments.json template which is in use by the cluster limits configurations.   I wasn't sure of the implications, but I can do it if needed - just let me know.

@wabouhamad @chaitanyaenr  PTAL

### DIFF
--- a/openshift_scalability/config/cluster-limits-deployments-per-namespace.yaml
+++ b/openshift_scalability/config/cluster-limits-deployments-per-namespace.yaml
@@ -5,7 +5,7 @@ projects:
     templates:
       - 
         num: 2000 
-        file: ./content/deployment-config-cluster-limits-deployments.json
+        file: ./content/deployment-cluster-limits-deployments.json
         parameters:
           -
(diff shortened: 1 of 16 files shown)


wabouhamad:
@mffiedler LGTM

chaitanyaenr:
@mffiedler Nice, let's modify the deployment config used by the cluster limits tests. We need to replace deploymentconfig templates in https://github.com/openshift-scale/workloads as well.

mffiedler:
> @mffiedler Nice, let's modify the deployment config used by the cluster limits tests. We need to replace deploymentconfig templates in https://github.com/openshift-scale/workloads as well.

Updated the cluster limit tempates and configs.  Also update the description of the PR to indicate dc are not officially deprecated.   Merging if there are no further comments.

PR #637, opened 2020-04-21
Storage mongodb test update

skordas:
From version 4.x whole test is running from client using only oc client commands - no ansible needed.


skordas:
@rpattath PTAL

PR #640, opened 2020-07-17
Avoid scheduling objects on infra nodes

chaitanyaenr:
The infra nodes run critical components including prometheus, logging,
router and registry. This commit will make sure the application pods
land on worker nodes instead of infra.


chaitanyaenr:
@mffiedler PTAL.

PR #641, opened 2020-07-20
fix: make build work

georgettica:


georgettica:
@mffiedler can you TAL at this? this makes it more explicit which python is required to run this

PR #644, opened 2020-08-27
Update taint/toleration test for 4.x

mffiedler:
Update for 3 m5.2xlarge worker nodes and 3 masters (test case in Polarion updated to match 4.x).

@wabouhamad PTAL .   Was there any reason to use absolute paths (/root/svt....) in the config file vs. relative paths?   It seems the calling wrapper (shell script, ansible, etc) could set the working directory and use relative paths.   Let me know if that causes problems.


wabouhamad:
@mffiedler no specific reasons for absolute paths, just making sure the file will always be accessible.  LGTM

mffiedler:
Thanks @wabouhamad

PR #649, opened 2020-12-10
Updating Mongo version

paigerube14:
For https://issues.redhat.com/browse/OCPQE-2712

This will update the mongodb version in the test-mongo-load.sh file to not need to hard code. It will be based on the number from the mongo_version variable in the external_vars.yaml file. If given latest in the original external_vars file I will get the number version to overwrite the file


storage/mongodb/mongodb.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+vars_file='external_vars.yaml'
+output_file="mongodb.out"
+rm $output_file
+mongodb_version=$(cat $vars_file | grep MONGODB_VERSION | cut -d ' ' -f 2)
+
+if [[ $mongodb_version == "latest" ]]; then
```
skordas, line 8: ohhh... I like this!

skordas: I would add information in README you can use the latest version - and maybe change that in external_vars file - we want to run against the latest one - if that is not some retest.

paigerube14: Updated readme to explain the latest variable in the external_vars. And pointed to run mongodb.sh instead of runmongo.sh

skordas (approved):

PR #651, opened 2021-01-05
Reduce reliability logging - remove pods in good state from output

mffiedler:
The reliability logs get huge over time, mainly due to the oc get pods --all-namespaces output.  Pods in Running or Complete status aren't interesting for the purpose of monitoring reliability runs, we really care about Error, CrashLoop, stuck in Terminating, etc.  

Reduce the logged output by changing this to oc get pods --all-namespaces | egrep -v "Complete|Running"


mffiedler:
@wabouhamad PTAL

PR #652, opened 2021-01-06
Pod and Node Affinity updates

paigerube14:
These changes are adding both pod and node affinity test scripts.

- Added python and golang versions of cluster loader
- Making relative imports in yaml files 
- Adding python yaml files for python cluster loader 
- Add validation of expected number of pods in both pods and nodes based on nums set in yaml files not hard coded 
- This also includes using the 4.6 version of openshift-tests 

Also adding a script that helps install the openshift-tests binary (get_openshift_tests.sh )

### DIFF
--- a/openshift_scalability/ci/scripts/README.md
+++ b/openshift_scalability/ci/scripts/README.md
@@ -1,0 +1,68 @@
+## Pod and Node Affinity README
+
+### Purpose 
+The pod and node affinity tests, test the ability of pods getting scheduled using Affinity and Anti-Affinity rules as we approach the capacity of compute nodes  
+
+
+### Use Python2 if running using python cluster loader
+
+The following command should return python 2
+```
+$ python --version 
+```
+
+If not, consider using a python 2 virtual environment.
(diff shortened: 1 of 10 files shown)


paigerube14:
@wabouhamad when you get a chance can you take a look and see if this what you were thinking about for some updates to these scripts? And also anything else I can add. Thanks!

mffiedler (approved):
LGTM - Some nitpick comments on the utility script.  But the real test changes are good - nice cleanup of these tests.

mffiedler:
Tested python and golang.

mffiedler:
The node test is passing for both python and golang.   The pod test passes for golang but fails for python.  Investigating.

paigerube14:
> The node test is passing for both python and golang. The pod test passes for golang but fails for python. Investigating.

I hit the same issue I think. I added an "s" to the pause time in the pod-affinity.yaml file. I am going to add a README because the cluster loader needs to use python2 to properly decode in the utils.py script.

mffiedler:
All tests working for me now...merging.

paigerube14:
Thanks for reviewing this @mffiedler, much appreciated

PR #653, opened 2021-01-08
Conc builds version updates and README

paigerube14:
This PR will update the versions of the imagestreamtag in both the django and cakephp build files.

Also including a README with some common errors and packages needed


mffiedler (approved):
/LGTM

PR #656, opened 2021-01-28
Statefulsets performance 

paigerube14:
This is adding a set automation script for the statefulsets performance [TC](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-15356)


paigerube14:
I used 6 worker nodes at m5.4xlarge size

mffiedler:
There's a stray empty file from pod_density in the commit.

paigerube14:
> There's a stray empty file from pod_density in the commit.

I think I need the init file to be able to use one of the functions in the pod_density folder. https://github.com/openshift/svt/blob/f9cc08ac7371ed01cc4a5bdf5a0500c8ab065288/openshift_scalability/scripts/statefulset_perf.sh#L10

mffiedler:
Got it - missed that.  Disregard comment.

openshift_scalability/scripts/statefulset_perf.sh
```diff
@@ -0,0 +1,95 @@
...
+  COUNTER=0
+  nodes=$(oc get nodes -l 'node-role.kubernetes.io/worker='  | awk '{print $1}'| grep -v NAME )
+  for node_name in $nodes
+  do
+    oc describe node $node_name
+
+  done
+}
+
+CONFIG_FILE=../config/pyconfigStatefulSet.yaml
```
mffiedler, line 70: Path math problem again.   This yaml file has ./content/statefulset-pv-template.json in it which results in the path "./content/statefulset-pv-template.json".    Easy fix is to update this yaml to be ../content

paigerube14: should be fixed now

PR #657, opened 2021-02-16
Error cases and logs for concurrent builds 

paigerube14:
This PR adds in more details to the readme for the concurrent builds script. Adding in possible debugging steps and 2 helpful solutions to common issues I've found.

It adds in a script that gets all the logs of failed builds and pods that are not in a "Running|Complete|Creating|Pending" state. I have found this super helpful when there are more than a few failed builds for opening bugs.

### DIFF
--- a/openshift_performance/ci/scripts/README.md
+++ b/openshift_performance/ci/scripts/README.md
@@ -1,3 +1,15 @@
+## Concurrent builds README
+
+### Purpose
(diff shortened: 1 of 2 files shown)


mffiedler (approved):
Doc changes LGTM and since the get_conc_build_info.py is net new, I am merging it.  If there are issues, we can address as a follow up

PR #658, opened 2021-02-17
Loaded Projects 

paigerube14:
Updating [TC](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-9461) to be able to use python cluster loader and separating out the cluster loader part to be able to easily run without pbench.

### DIFF
--- a/openshift_scalability/ci/scripts/create-loaded-projects.sh
+++ b/openshift_scalability/ci/scripts/create-loaded-projects.sh
@@ -1,0 +1,120 @@
+#!/bin/bash 
+
+if [ "$#" -ne 1 ]; then
+  echo "syntax: $0 <TYPE>"
+  echo "<TYPE> should be either golang or python"
+  exit 1
+fi
+
+TYPE=$1
+
+function python_clusterloader() {
+  MY_CONFIG=../../config/configCreateLoadedProjects.yaml
+  python_v=$(python --version 2>&1 | sed 's/.* \([0-9]\).\([0-9]\).*/\1\2/')
+  if [[ $python_v -lt "29" ]]; then
+    echo "Running python 2";
+  else
+    echo "Have python version: $(python --version 2>&1); Need to run with python 2: exiting"
+    exit
+  fi
+
+  python ../../cluster-loader.py -f $MY_CONFIG
+}
+
+function golang_clusterloader() {
+
+  # start GoLang cluster-loader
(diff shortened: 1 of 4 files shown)


openshift_scalability/ci/scripts/create-loaded-projects.sh
```diff
@@ -0,0 +1,120 @@
...
+
+function golang_clusterloader() {
+
+  # start GoLang cluster-loader
+  export KUBECONFIG=${KUBECONFIG-$HOME/.kube/config}
+  cur_loc=$(pwd)
+
+  MY_CONFIG=$cur_loc/../../config/golang/configCreateLoadedProjects.yaml
+
+  VIPERCONFIG=$MY_CONFIG openshift-tests run-test "[sig-scalability][Feature:Performance] Load cluster should populate the cluster [Slow][Serial]"
```
mffiedler, line 32: Given the relative paths inside of config/golang/configCreateLoadedProjects.yaml this failed for me when I ran it from ci/scripts directory.   Which has been a weak point of golang CL and our directory structure.   I had to add the following line above the call to openshift-tests:
  cd ../../content
to get the ./<template-name> templates in configCreateLoadedProjects.yaml to work. 

We can either update the script or the yaml file, but I think we should be able to run the script from ci/scripts.

mffiedler (changes requested):

openshift_scalability/config/golang/configCreateLoadedProjects.yaml
```diff
@@ -15,19 +15,19 @@ ClusterLoader:
...
-          file: ./image-stream-all-template.json
+          file: ../../image-stream-all-template.json
         -
           num: 200
-          file: ./ssh-secret-template.json
+          file: ../../ssh-secret-template.json
         -
           num: 10
-          file: ./route-template.json
+          file: ../../route-template.json
```
mffiedler, line 30: These should all be under content

paigerube14:
Updated pointers to templates in golang config.

Having an issue that the objects created by the templates (not pods) are not being created in the project, they are getting created in the default project. Still investigating this issue

mffiedler (commented):
In general, LGTM.  Please squash.  Curious as to why you picked lower default project and pod numbers?

mffiedler (approved):
Tested python and golang - all working well.   LGTM

paigerube14:
Thanks Mike! Sorry this took so long to make updates.

PR #659, opened 2021-02-17
Node vertical updates

paigerube14:
Updating the node vertical test to have non hard coded paths and updated to work with the golang cluster loader version 4.7

### DIFF
--- a/openshift_scalability/config/golang/nodeVertical-labeled-nodes.yaml
+++ b/openshift_scalability/config/golang/nodeVertical-labeled-nodes.yaml
@@ -7,10 +7,9 @@ ClusterLoader:
       tuning: default
       ifexists: delete
       pods:
-        - num: 1000
-          image: gcr.io/google_containers/pause-amd64:3.0
-          basename: pausepods
-          file: pod-pause-labeled-nodes.json
+        - num: 660
+          basename: hello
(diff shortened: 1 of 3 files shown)


paigerube14:
Updated configs to be 660 pods (and test case to have 3 workers) 

Should I update the config file to point to a file with a hello image instead of the pod-pause.json?

openshift_scalability/config/golang/nodeVertical-labeled-nodes.yaml
```diff
@@ -7,10 +7,10 @@ ClusterLoader:
...
       ifexists: delete
       pods:
-        - num: 1000
-          image: gcr.io/google_containers/pause-amd64:3.0
-          basename: pausepods
-          file: pod-pause-labeled-nodes.json
+        - num: 660
+          image: openshift/hello-openshift
+          basename: hello
+          file: ../../content/pod-pause-labeled-nodes.json
```
mffiedler, line 13: We want to use pause pods for this test because they are the smallest possible container to run that we have.   They make the test purely a test of the control plane and node infrastructure - no user or OS code involved.

paigerube14:
Took out the image name from the config yaml files for both python and golang versions. It was not being used and doesn't cause issues taking it out

mffiedler (approved):
LGTM - thanks for the updates.

PR #664, opened 2021-04-22
Pod density updates

paigerube14:
Updating the pod density script to:

- have 22 worker nodes (20 schedulable and 2 cordoned) 
- Cordons the 2 nodes that have prometheus pods running on them
- Rewrote the way to get the counts of the number of pods on a node 
- Line 107 and 112 in the bash script had to be updated because I had a multus pod that contained the string "svt" and was causing failures in run


mffiedler (approved):
Tested on AWS - LGTM

PR #665, opened 2021-04-27
Pvc delete automation script

paigerube14:
Created an automation script for [pvc delete test case](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-23927). It will print out the deletion time of each of the pvc nums and iterations into a file called pvc_delete.out. It will create the sk2 storage class if it is not already created.

Top parameters to edit
- Number of pvc's to create (using cluster loader)
- Number of iterations you want to execute 

The init files are need to be able to reuse the rewriting of the yaml files based on the different pvc numbers we want to create in each iteration

### DIFF
--- a/openshift_scalability/pvc_delete.sh
+++ b/openshift_scalability/pvc_delete.sh
@@ -1,0 +1,81 @@
+#/!/bin/bash
+
+pvc_nums=(10 20 30 50 100 200)
+iterations_num=3
+results_file=pvc_delete.out
+loader_file=./content/pvc-templates/pvc-parameters.yaml
+cloud_provider=aws
+
+function create_projects() {
+  echo "create projects variable $loader_file"
+  python cluster-loader.py -f $loader_file -v
+}
+
+function rewrite_yaml() {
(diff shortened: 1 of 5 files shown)


paigerube14:
@rpattath are you able to take a look at this PR. Saw you have been running this test case the most recently

rpattath (approved):
LGTM

openshift_scalability/pvc_delete.sh
```diff
@@ -0,0 +1,81 @@
+#/!/bin/bash
```
pruan-rht, line 1: @paigerube14 wondering why you decided to use bash shell inside a python directory instead of creating a binding python script instead.

paigerube14: For the most part I've seen that the scripts in this repo are in bash. I also was incorporating a smaller pre-existing bash script into this one to create a fully encompassed test case.
I could create it in python. Thoughts?

pruan-rht: @paigerube14 others can chime in.  IMO it's just harder to debug with bash when something goes wrong (could just be my bash-fu is horrible).

PR #668, opened 2021-05-27
add personas and kuebconfig with context for multiple users

qiliRedHat:
https://issues.redhat.com/browse/OCPQE-4095
1. Add 'persona' and 'concurrency' entries to the config file, in task section, to specify which kind of user this task will be executed by, and how many of users will do this task concurrently.
config file:
```
  tasks:
    minute:
      - action: create
        resource: projects
        quantity: 1
        persona: developer
        concurrency: 2
```
3. To make multiple user to run oc command concurrently, make a copy of kubeconfig file provided by installer for each user, and addd context for each user in each kubeconfig file by doing a login with that user.
Add 'kubeconfig' in config file pointing to the kubeconfig get from the cluster install.
config file:
```
reliability:
  kubeconfig: ~/Downloads/kubeconfig
```

Add a module 'Context' to create the kubeconfig files (copy from the kubeconfig in config file as above) and add contexts.

example context in 'kubeconfig_testuser-0' after login with 'testuser-0'
```
contexts:
- context:
    cluster: api-qili-qe-devcluster-openshift-com:6443
    user: testuser-0/api-qili-qe-devcluster-openshift-com:6443
  name: /api-qili-qe-devcluster-openshift-com:6443/testuser-0
current-context: /api-qili-qe-devcluster-openshift-com:6443/testuser-0
```
The creation of kubeconfig and context will be initiated once before test start. Installer (Flexy install) provides 1 admin users kubeadmin and 50 developer users, testuser-0 to testuser-49. Create kubeadmin user's context directly. Create the 50 developer uesrs' context with multi thread to speed up the initiation work.

3. As the kubeconfigs (and sometimes users) data will not be changed after init and need to be used by all tasks, for convenience, create a singleton  module 'GlobalData' to hold those data. To be consistent, put the config data into this instance too. Modify related modules as needed to adapt to this config data change.

Exit Criteria
1. Reliability test creates 1 admin and 50 developer users' kubeconfig file with context right after test started and before executing tasks. The kubeconfig files are put under a 'kubeconfigs' folder in the same location of the reliability.py. 
2. Edit any previous config files to add 'kubeconfig' and run with it, test should pass.
Add 'kubeconfig' in your config file and point it to the kubeconfig file downloaded from installer.
example:
```
reliability:
  kubeconfig: ~/Downloads/kubeconfig
```
3. Run test with 'enhance_reliability.yaml' to test the multiple users login with kubeconfig files, should pass.

### DIFF
--- a/reliability/config/enhance_reliability.yaml
+++ b/reliability/config/enhance_reliability.yaml
@@ -1,0 +1,40 @@
+reliability:
+  kubeconfig: ~/Downloads/kubeconfig
+  timeSubstitutions:
+    minute: 10s
+    hour: 30s
+    day: 2m
+    week: 3m
+    month: 4m
+  limits:
+    maxProjects: 20
+    # sleepTime should be <= mix of timeSubstitutions in the current design,
+    # or else, tasks with timeSubstitutions < sleepTime will be executed with the interval of sleepTime.
+    sleepTime: 10
+
+  appTemplates:
+    - template: cakephp-mysql-persistent
+    - template: nodejs-postgresql-persistent
+    - template: django-psql-persistent
+    - template: rails-pgsql-persistent
+    - template: dancer-mysql-persistent
+
+  users:
+    - admin_user: kubeadmin
+    - admin_password: R3xrm-uIABR-BB4Za-xaWWF
+    - user_file: /Users/qili/Downloads/users.spec
+  tasks:
+    minute:
+      # - action: login
+      #   resource: session
+      #   persona: admin
+      #   concurrency: 1
+      - action: login
+        resource: session
+        persona: developer
+        concurrency: 3
+      # - action: create
+      #   resource: projects
+      #   quantity: 1
+      #   persona: developer
+      #   concurrency: 2
--- a/reliability/tasks/GlobalData.py
+++ b/reliability/tasks/GlobalData.py
@@ -1,0 +1,44 @@
+from ReliabilityConfig import ReliabilityConfig
+from tasks.Users import all_users, User
+from tasks.Contexts import all_contexts
+import logging
+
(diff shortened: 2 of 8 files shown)


qiliRedHat:

reliability/config/enhance_reliability.yaml
```diff
@@ -0,0 +1,40 @@
...
+    minute:
+      # - action: login
+      #   resource: session
+      #   persona: admin
+      #   concurrency: 1
+      - action: login
+        resource: session
+        persona: developer
+        concurrency: 3
+      # - action: create
```
mffiedler, line 36: Let's at least have project creation in the sample

reliability/tasks/GlobalData.py
```diff
@@ -0,0 +1,44 @@
+from ReliabilityConfig import ReliabilityConfig
+from tasks.Users import all_users, User
+from tasks.Contexts import all_contexts
+import logging
+
+# a class to hold shared data, config, users, kubeconfigs
+class GlobalData:
```
mffiedler, line 7: better approach - thanks.

mffiedler (changes requested):

reliability/config/enhance_reliability.yaml
qiliRedHat: Don't worry, I only put session - long in  in this PR's sample,  because the Projects/Apps/Pods modules are not yet multiple users ready, however one can still include project creation but that will still be the old way. After this is merged, in the phase2 - adding concurrency PR, we can have all kinds of tasks ready for multiple users.

reliability/tasks/GlobalData.py
qiliRedHat: Thanks!

PR #669, opened 2021-06-04
Bulk delete

paigerube14:
Full script for [TC](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-14475)

This test case takes a very very long time to run. For testing might be easier to reduce numbers. 

I edited the delete_projects script so that the output of deleting each of the projects goes to a file and the timing will be put to stdout, let me know if I should change this. 

For the correct spacing when rewriting I had to use a python3 pip package but cluster loader needs python2 currently. So had to specify the python version when running the rewrite script. Let me know if you don't think this will work or if there's a better way

### DIFF
--- a/openshift_performance/ci/scripts/bulk_delete.sh
+++ b/openshift_performance/ci/scripts/bulk_delete.sh
@@ -1,0 +1,65 @@
+#/!/bin/bash
+#set -x
+################################################
+## Auth=prubenda@redhat.com
+## Desription: Script for running bulk delete
+## of empty and loaded projects and comparing times
+################################################
+file_1="../content/conc_proj.yaml"
+file_2="../content/pyconfigLoadedProject.yaml"
+delete_output=bulk_delete.out
+function create_projects()
+{
+  python ../../../openshift_scalability/cluster-loader.py -f $1
+}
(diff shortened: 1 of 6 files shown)


paigerube14:
Added a comparison script based on the output from bulk_delete.sh

openshift_performance/ci/scripts/bulk_delete.sh
```diff
@@ -0,0 +1,67 @@
...
+
+function rewrite_yaml() {
+  python3 -c "import increase_pods; increase_pods.print_new_yaml($1,'$2')"
+}
+
+python3 -m pip install ruamel.yaml
+
+rm -if $delete_output
+
+exit 0
```
mffiedler, line 28: This halts the script.   Debug artifact?

paigerube14: yes definitely not supposed to be there....deleting

mffiedler:
PR LGTM.  I did hit an issue where oc delete project -l purpose=test left 20 projects behind - not terminating, but active.   I'll investigate that.

PR #670, opened 2021-06-10
add concurrency to reliability test

qiliRedHat:
https://issues.redhat.com/browse/OCPQE-4221

### Concurrent task run with multiple users - priority 0-1
**What**
Today the test execution is sequentially with default user system:admin and then as the user specified in the "login" action.
This task is to add concurrency for each type of user added in [phase 1](https://issues.redhat.com/browse/OCPQE-4095).
X admin/developer users can run m tasks(oc commands) concurrently. (currently only one admin user in config yaml structure, can restructure in the future)
Y end users can visit n applications concurrently

**Why**
1. Simulate concurrent operations from different users like the real world.
2. Shorten the time of test execution.
For example, for 51 users doing the login operation:
sequentially: 243.6s
concurrently(51 threads): 13.0s

**How**
1. According to the setting of task in config file, choose 'concurrency' number of users for each 'persona' (admin, developer, user). With the config below, 'testuser-0' and 'testuser-1' will execute the project create operation concurrently, while 10 end users will visit all (100%) applications - every application will be visited by 10 users concurrently.
```
      - action: create
        resource: projects
        quantity: 2
        persona: developer
        concurrency: 2
      - action: visit
        resource: apps
        applyPercent: 100
        persona: user
        concurrency: 10
```
2. Modify Task.py, for each type of action, use multiple threads to run the tasks with oc or shell commands.

3. Modify functions Projects.py, Apps.py, and Pods.py to accept 'kubeconfig' parameter. 
Add -n namespace to a command if needed, such as getting pod - a dev user can only list pods under its own namespace.
For operations that can only be run by admin, such as labeling namespace, using admin user's kubeconfig.

4. Add global locks to protect the shared mutable counters, like the project id, total project number, app number, build number.
5. The '+=' operations are not thread safe. 
Is adding and poping a dict's entry in a shared map with multiple thread thread safe? - I think, yes. https://docs.python.org/3/glossary.html#term-global-interpreter-lock
https://docs.python.org/3/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe

5. Handle the situation caused by ansync operations. For example, an app is chosen to be upgraded, while the project deletion (--wait=false is async) actually done after that.

6. For app visit, the current solution is visiting sequentially. Implement concurrency with corutines (asyncio|https://github.com/timofurrer/awesome-asyncio/blob/master/README.md + aiohttp.
Corutines can support high concurrency for persona user. Concurrency is better than multi threading as it is single-threaded, multiplexing I/O access, thus cost is lower.
Add a new LoadApp module to load apps, use this module to count the application visit succeeded and failed numbers. 
The visit function is Apps module is kept for verify application route availability after app creation. 
In this way, the app visit succeeded and failed numbers only count the visits after route is available, 503 before available will not be counted. This can better reflect the user persona's concern - app visiting after it is ready.
example of an app visit task config. This config means 10 persona user will visit all applications concurrently.
```
      - action: visit
        resource: apps
        applyPercent: 100
        persona: user # actually not used
        concurrency: 10
```

**Exit Criteria**
Config with all kinds of tasks can be run with multiple users in parallel.
`python reliability.py -f config/enhance_reliability.yaml`

### DIFF
--- a/reliability/config/enhance_reliability.yaml
+++ b/reliability/config/enhance_reliability.yaml
@@ -1,15 +1,16 @@
 reliability:
-  kubeconfig: ~/Downloads/kubeconfig
+  kubeconfig: <path_to_kubeconfig>
   timeSubstitutions:
-    minute: 10s
-    hour: 30s
+    minute: 30s
+    hour: 1m
     day: 2m
-    week: 3m
-    month: 4m
+    week: 6m
+    month: 6m
   limits:
-    maxProjects: 20
-    # sleepTime should be <= mix of timeSubstitutions in the current design,
-    # or else, tasks with timeSubstitutions < sleepTime will be executed with the interval of sleepTime.
+    # total number of projects to create
+    # for 3 nodes m5.xlarge cluster, 30 is recomended
+    # for 5 nodes m5.xlarge cluster, 60 is recomended
+    maxProjects: 25
     sleepTime: 10
 
   appTemplates:
@@ -18,23 +19,63 @@ reliability:
     - template: django-psql-persistent
     - template: rails-pgsql-persistent
     - template: dancer-mysql-persistent
-
   users:
-    - admin_user: kubeadmin
-    - admin_password: <password>
-    - user_file: /Users/qili/Downloads/users.spec
+    - kubeadmin_file: <path_to_kubeadmin-password>
+    - user_file: <path_to_users.spec>
   tasks:
     minute:
-      # - action: login
-      #   resource: session
-      #   persona: admin
-      #   concurrency: 1
+      - action: check
+        resource: pods
+        persona: admin
+        concurrency: 1
+      - action: check
+        resource: projects
+        persona: developer
+        concurrency: 5
+    hour:
+      - action: check
+        resource: projects
+        persona: developer
+        concurrency: 5
+      - action: create
+        resource: projects
+        quantity: 2
+        persona: developer
+        concurrency: 5
+      - action: visit
+        resource: apps
+        applyPercent: 100
+        persona: user
+        concurrency: 10
+      - action: scaleUp
+        resource: apps
+        applyPercent: 50
+        persona: developer
+        concurrency: 3
+      - action: scaleDown
+        resource: apps
+        applyPercent: 50
+        persona: developer
+        concurrency: 1
+      - action: build
+        resource: apps
+        applyPercent: 33
+        persona: developer
+        concurrency: 2
+      - action: modify
+        resource: projects
+        applyPercent: 25
+        persona: developer
+        concurrency: 2
+      - action: clusteroperators
+        resource: monitor
+    week:
+      - action: delete
+        resource: projects
+        applyPercent: 33
+        persona: developer
+        concurrency: 5
       - action: login
         resource: session
         persona: developer
-        concurrency: 3
-      # - action: create
-      #   resource: projects
-      #   quantity: 1
-      #   persona: developer
-      #   concurrency: 2
+        concurrency: 5
--- a/reliability/tasks/Apps.py
+++ b/reliability/tasks/Apps.py
@@ -1,3 +1,4 @@
+from .GlobalData import global_data
 from  .utils.oc import oc
 import requests
 import time
@@ -11,17 +12,17 @@ class App:
         self.deployment = deployment
         self.build_config = build_config
         self.route = route
-        self.build_count = 0
-        self.visit_succeded = 0
-        self.visit_failed = 0
         self.logger = logging.getLogger('reliability')
 
-    def build(self):
-        (result, rc) = oc("start-build -n " + self.project + " " + self.build_config)
+    def build(self, kubeconfig):
+        (result, rc) = oc("start-build -n " + self.project + " " + self.build_config, kubeconfig)
         if rc != 0:
             self.logger.error("build_app: Failed to create app " + self.deployment + " in project " + self.project)
+            return "App build failed for build config : " + self.build_config
         else:
-            self.build_count += 1
+            with global_data.builds_lock:
+                global_data.total_build_count += 1
+            return "App build succeeded for build config : " + self.build_config
 
     def visit(self):
         visit_success = False
@@ -29,51 +30,47 @@ class App:
             r = requests.get("http://" + self.route + "/")
             self.logger.info(str(r.status_code) + ": visit: " + self.route)
             if r.status_code == 200:
-                self.visit_succeded += 1
                 visit_success = True
-            else:
-                self.visit_failed += 1
-        except Exception:
-            self.visit_failed += 1
-            self.logger.error("visit: " + self.route)
+        except Exception as e :
+            self.logger.error(f"visit: {self.route} Exception {e}")
         return visit_success
-        
 
-    def scale_up(self):
-        (result, rc) = oc("scale --replicas=2 -n " + self.project + " dc/" + self.deployment)
+    def scale_up(self, kubeconfig):
+        (result, rc) = oc("scale --replicas=2 -n " + self.project + " dc/" + self.deployment, kubeconfig)
         if rc !=0 :
             self.logger.error("scale_up: Failed to scale up " + self.project + "." + self.deployment)
+            return "App scale up failed for deployment : " + self.deployment
+        else:
+            return "App scale up succeeded for deployment : " + self.deployment
     
-    def scale_down(self):
-        (result, rc) = oc("scale --replicas=1 -n " + self.project + " dc/" + self.deployment)
+    def scale_down(self, kubeconfig):
+        (result, rc) = oc("scale --replicas=1 -n " + self.project + " dc/" + self.deployment, kubeconfig)
         if rc !=0 :
             self.logger.error("scale_down: Failed to scale down " + self.project + "." + self.deployment)
-        
+            return "App scale down failed for deployment : " + self.deployment
+        else:
+            return "App scale down succeeded for deployment : " + self.deployment
     
 class Apps:
     def __init__(self):
-        self.build_count = 0
-        self.visit_succeded = 0
-        self.visit_failed = 0
         self.failed_apps = 0
         self.apps = {}
         self.logger = logging.getLogger('reliability')
 
-    def add(self, app):
-        (result, rc) = oc("new-app -n " + app.project + " --template " + app.template)
+    def add(self, app, kubeconfig):
+        (result, rc) = oc("new-app -n " + app.project + " --template " + app.template, kubeconfig)
         if rc != 0:
             self.logger.error("create_app: Failed to create app " + app.deployment + " in project " + app.project)
+            return None
         else:
             self.apps[app.project + "." + app.deployment] = app
-            (route,rc) = oc("get route --no-headers -n " + app.project + " | awk {'print $2'} | grep " + app.template)
+            (route,rc) = oc("get route --no-headers -n " + app.project + " | awk {'print $2'} | grep " + app.template, kubeconfig)
             if rc == 0:
                 app.route = route.rstrip()
                 max_tries = 60
                 current_tries = 0
                 visit_success = False
                 while not visit_success and current_tries <= max_tries:
-                    # don't count failed visits before app is ready
-                    app.visit_failed = 0
                     self.logger.info(app.template + " route not available yet, sleeping 10 seconds") 
                     time.sleep(10)
                     current_tries += 1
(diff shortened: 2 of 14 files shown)


qiliRedHat:

reliability/tasks/Apps.py
```diff
@@ -81,19 +84,19 @@ def add(self, app):
...
+        self.build_count = 0
         for key in self.apps.keys():
             self.build_count += self.apps[key].build_count
-            self.visit_succeded += self.apps[key].visit_succeded
-            self.visit_failed += self.apps[key].visit_failed
 
-    
     # removing an app just removes the dictionary entry, actual app removed by project deletion
     def remove(self,app):
+        # Q: why adding build_count when removing app?
```
mffiedler, line 98: I think the intention was to make sure any builds that happened since self.build_count was updated were added to the counter. But, it may be a bug.

mffiedler:
In my first test, after 33 minutes I got the following stack - looks like it was first time project deletion was run:
2021-06-11 20:36:10,038 - INFO - Project deleted : testuser-3-2
2021-06-11 20:36:10,042 - INFO - 0: project.project.openshift.io "testuser-4-1" deleted

2021-06-11 20:36:10,042 - INFO - Project deleted : testuser-4-1
Traceback (most recent call last):
  File "/root/tmp/svt/reliability/reliability.py", line 28, in <module>
    task_manager.start()
  File "/root/tmp/svt/reliability/tasks/TaskManager.py", line 134, in start
    task.execute()
  File "/root/tmp/svt/reliability/tasks/Task.py", line 254, in execute
    login_args.append((name, password, kubeconf))
NameError: name 'name' is not defined

Have not investigated yet.

reliability/tasks/Apps.py
qiliRedHat: I got it. The total builds in the summary is the total builds ever done during the test, including the builds of the removed apps. Then I think the logic of func refresh_stats is not correct, because it gets all the build numbers of the current existing apps. https://github.com/openshift/svt/blob/fb572b272f3e897eeac1c1b8d45c1dc7fd5ff683/reliability/tasks/Apps.py#L89-L94

I'll modify this counter.

qiliRedHat:
> In my first test, after 33 minutes I got the following stack - looks like it was first time project deletion was run:
> 2021-06-11 20:36:10,038 - INFO - Project deleted : testuser-3-2
> 2021-06-11 20:36:10,042 - INFO - 0: project.project.openshift.io "testuser-4-1" deleted
> 
> 2021-06-11 20:36:10,042 - INFO - Project deleted : testuser-4-1
> Traceback (most recent call last):
> File "/root/tmp/svt/reliability/reliability.py", line 28, in
> task_manager.start()
> File "/root/tmp/svt/reliability/tasks/TaskManager.py", line 134, in start
> task.execute()
> File "/root/tmp/svt/reliability/tasks/Task.py", line 254, in execute
> login_args.append((name, password, kubeconf))
> NameError: name 'name' is not defined
> 
> Have not investigated yet.

This is not about project deletion, but about login action after project deletion. This is a bug introduced when I update the login part to make it to use the new login function in Session.py. I fixed it. Thanks!

reliability/tasks/Apps.py
qiliRedHat: Modified build counter in commit https://github.com/openshift/svt/pull/670/commits/1a854b15f85ce36b7b0f45a6fcd42dc511d73e9b

qiliRedHat:
@mffiedler Thank you for the first review. Please check my reply and comment above. After the review is done, I'll do a rebase to keep a single commit.

reliability/tasks/Task.py
```diff
@@ -101,45 +152,112 @@ def execute(self):
...
+                    app_build_args = []
                     for app_to_build in apps_to_build:
-                        all_apps.apps[app_to_build].build()
+                        try:
+                            project = all_apps.apps[app_to_build].project
+                            user = all_projects.projects[project].user
+                            kubeconfig = global_data.kubeconfigs[user]
+                            app_build_args.append((app_to_build, kubeconfig))
+                        except KeyError:
+                            self.logger.info(f"Projet {project_to_modify} no longer exists in projects list.")
```
paigerube14, line 164: Don't think the project_to_modify variable will be set here, need to change to just _project_ or change the logging line. 
Missing **c** in project as well at the beginning of the line

reliability/tasks/Task.py
```diff
@@ -101,45 +152,112 @@ def execute(self):
...
+                    # prepare parameters list - app name and kubeconfig - for app.scale_up
+                    app_scale_args = []
+                    for app_to_scale in apps_to_scale:
+                        try:
+                            project = all_apps.apps[app_to_scale].project
+                            user = all_projects.projects[project].user
+                            kubeconfig = global_data.kubeconfigs[user]
+                            app_scale_args.append((app_to_scale, kubeconfig))
+                        except KeyError:
+                            self.logger.info(f"Projet {project_to_modify} no longer exists in projects list.")
```
paigerube14, line 185: Same logging issues here: don't think the project_to_modify variable will be set here, need to change to just project or change the logging line.

paigerube14:
```
    result = self.fn(*self.args, **self.kwargs)
  File ".../svt/reliability/tasks/Task.py", line 243, in <lambda>
    results = executor.map(lambda t: all_pods.check(*t), pod_check_args)
  File ".../svt/reliability/tasks/Pods.py", line 10, in check
    if user == "kubeadmin":
NameError: name 'user' is not defined
```

Qiujie, I am getting this error when running. I think it is coming from inside the check pods function if statement around line [240](https://github.com/openshift/svt/blob/1a854b15f85ce36b7b0f45a6fcd42dc511d73e9b/reliability/tasks/Task.py#L240) that you are missing passing the user


After adding the user to that line I get the following issue. It seems that when checking for the pods it's not looking in the correct project, you're only looking in the default project. Probably just need to add "-n <project_name>" to the line when you are getting the pods. 
```
2021-06-15 09:54:49,933 - INFO - =>oc get pods --kubeconfig /Users/prubenda/PycharmProjects/pplRepos/qili/svt/reliability/kubeconfigs/kubeconfig_kubeadmin| egrep -v "Running|Complete"
2021-06-15 09:54:49,933 - INFO - Pod check failed for namespace : kubeadmin-4
2021-06-15 09:54:50,355 - INFO - 1: No resources found in default namespace.

2021-06-15 09:54:50,356 - ERROR - get pods: failed
```

Think you just need to have line [11](https://github.com/openshift/svt/blob/1a854b15f85ce36b7b0f45a6fcd42dc511d73e9b/reliability/tasks/Pods.py#L11) match  the same format as line [13](https://github.com/openshift/svt/blob/1a854b15f85ce36b7b0f45a6fcd42dc511d73e9b/reliability/tasks/Pods.py#L13)

In that case do you need that if statement because they are the same line?

reliability/config/enhance_reliability.yaml
```diff
@@ -1,15 +1,13 @@
...
     day: 2m
-    week: 3m
-    month: 4m
+    week: 6m
+    month: 6m
   limits:
-    maxProjects: 20
-    # sleepTime should be <= mix of timeSubstitutions in the current design,
-    # or else, tasks with timeSubstitutions < sleepTime will be executed with the interval of sleepTime.
+    maxProjects: 25
```
mffiedler, line 10: It looks like with concurrency this has become a "projects per user number".  I am seeing 100+ projects active after a while which is more than a "standard" 3 worker cluster can handle.   We should change this number to 5 or 6 or else make it a global max.   Either is fine with me as long as we document it with a comment.

reliability/tasks/Task.py
qiliRedHat: Thank you for the catching!

qiliRedHat: Thank you for the catching!

reliability/tasks/Task.py
qiliRedHat: Thank you for the catching!

qiliRedHat:
> ```
>     result = self.fn(*self.args, **self.kwargs)
>   File ".../svt/reliability/tasks/Task.py", line 243, in <lambda>
>     results = executor.map(lambda t: all_pods.check(*t), pod_check_args)
>   File ".../svt/reliability/tasks/Pods.py", line 10, in check
>     if user == "kubeadmin":
> NameError: name 'user' is not defined
> ```
> 
> Qiujie, I am getting this error when running. I think it is coming from inside the check pods function if statement around line [240](https://github.com/openshift/svt/blob/1a854b15f85ce36b7b0f45a6fcd42dc511d73e9b/reliability/tasks/Task.py#L240) that you are missing passing the user
> 
> After adding the user to that line I get the following issue. It seems that when checking for the pods it's not looking in the correct project, you're only looking in the default project. Probably just need to add "-n <project_name>" to the line when you are getting the pods.
> 
> ```
> 2021-06-15 09:54:49,933 - INFO - =>oc get pods --kubeconfig /Users/prubenda/PycharmProjects/pplRepos/qili/svt/reliability/kubeconfigs/kubeconfig_kubeadmin| egrep -v "Running|Complete"
> 2021-06-15 09:54:49,933 - INFO - Pod check failed for namespace : kubeadmin-4
> 2021-06-15 09:54:50,355 - INFO - 1: No resources found in default namespace.
> 
> 2021-06-15 09:54:50,356 - ERROR - get pods: failed
> ```
> 
> Think you just need to have line [11](https://github.com/openshift/svt/blob/1a854b15f85ce36b7b0f45a6fcd42dc511d73e9b/reliability/tasks/Pods.py#L11) match the same format as line [13](https://github.com/openshift/svt/blob/1a854b15f85ce36b7b0f45a6fcd42dc511d73e9b/reliability/tasks/Pods.py#L13)
> 
> In that case do you need that if statement because they are the same line?

Thank you for catching this! 
Regarding the if, my intention was to let the admin persona user to show all pods in all namespaces, that's more like a real admin wants to do. You make me realized I need to add '-A' to show all namespaces's pods. 
I also need to investigate why 'ERROR - get pods: failed' is logged, because even though no resources found, the rc should be 0.
```
% oc get pods
No resources found in default namespace.
% echo $?
0
```

reliability/config/enhance_reliability.yaml
qiliRedHat: Yes, this is a per project now. Let me change the default number and then put some description to it, about how to set a good number considering the concurrency and the capacity of the cluster.

qiliRedHat: Jenkins is not working today, I used a manual installed aws cluster which does not have the 50 test users. So I didn't test multiple concurrent users yet. I added some debug output and did a test with admin user - concurrency=1, I can get the warning 
```
current_projects:25
max_projects:25
2021-06-16 19:58:18,830 - WARNING - create_project: project kubeadmin-796. 25 already exist.
```
I will test with real multiple users later.
I guess there are script to create the 50 test users for a manually installed cluster but haven't get chance to fine it, I'll figure out. Let me know if you know where it is.

mffiedler:
@qiliRedHat Can you squash the commits and then @paigerube14 and I can do final review/test?   Thanks.

reliability/config/enhance_reliability.yaml
```diff
@@ -18,23 +19,64 @@ reliability:
     - template: django-psql-persistent
     - template: rails-pgsql-persistent
     - template: dancer-mysql-persistent
-
   users:
     - admin_user: kubeadmin
     - admin_password: <password>
     - user_file: /Users/qili/Downloads/users.spec
```
paigerube14, line 25: can we make this similar to the password, to be more generic

ex.) <user.spec>

qiliRedHat:
> @qiliRedHat Can you squash the commits and then @paigerube14 and I can do final review/test? Thanks.

@mffiedler @paigerube14, I squashed the commits.

reliability/config/enhance_reliability.yaml
qiliRedHat: Thanks for the comment Paige. Fixed in the latest commit.

mffiedler:
One last change and I think this is good to go.   All users, especially kubeadmin, have to login once every 24 hours real time to keep their sessions alive.   Running over a weekend I saw admin oc commands start to fail with:

2021-06-21 02:00:44,373 - INFO - =>oc get clusteroperators --kubeconfig /root/tmp/svt2/reliability/kubeconfigs/kubeconfig_kubeadmin -o yaml
2021-06-21 02:00:44,643 - INFO - 1: apiVersion: v1
items: []
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
error: You must be logged in to the server (Unauthorized)

2021-06-21 02:00:44,643 - ERROR - get clusteroperators: failed
2021-06-21 02:01:14,673 - INFO - =>oc get pods -A --kubeconfig /root/tmp/svt2/reliability/kubeconfigs/kubeconfig_kubeadmin| egrep -v "Running|Complete"
2021-06-21 02:01:14,955 - INFO - 1: error: You must be logged in to the server (Unauthorized)


In the past this was handled by making the admin user login in a task, but we could make it part of the test infrastructure as well.
@qiliRedHat Let me know if you prefer to handle this as a separate issue after merging this PR.  I am ok with that too

qiliRedHat:
@mffiedler 
Great thanks to let me know this expiration.
https://github.com/openshift/svt/pull/670/commits/86059b3ed081620e8898a23621a02ed47adabc25 added the mechanism, to do re login in for all users  before a task run if it's been 23 hours since last login. I tested it with shorter time interval - 10 minutes, it works well.

mffiedler:
excellent.  Let me do one final test for 24+ hours and then I will plan on merging tomorrow.

mffiedler (approved):
Code LGTM and test has been running successfully against 4.8 3 days and has found 1 bug already :-)

@qiliRedHat - one more squash and I will merge it.

qiliRedHat:
@mffiedler Thanks you for the test and bug caching! I squashed.

paigerube14:
Reran yesterday, not for as long as Mike's test, things looked good to me!

qiliRedHat:
@paigerube14 Thank you for your time reviewing this pr and the bug you found!

PR #676, opened 2021-06-24
[reliability][enhance]add customized task

qiliRedHat:
https://issues.redhat.com/browse/OCPQE-4292

**What**
Add customized task type.

**Why**
In the current design the tasks types are fixed/hard coded and limited, for example, project creation/deletion/modify, app creation/build/scale up/scale/down/visit. Most of them belongs to developers and end users personas. However, there are more operations can be added to reliability tests. To simulate the real world's cluster operation, we need to add more task types. It's flexible to have customized type to allow user to specify any oc commandsor file contains oc commands as a task.

**How**
Add a new task type for customized oc command or file contains oc commands.
example of the config file piece:
```
  tasks:
    minute:
      # Specify an oc command to execute as 'action'.
      # Don't use command that could return '1' as expected, e.g. oc get pods -A | egrep -v "Running|Completed".
      # Use oc get pods -A | awk '$4!="Running" && $4!="Completed"' instead.
      - action: oc whoami
        resource: customize
        persona: developer
        concurrency: 5
      # Specify a file to execute as 'action'.
      # File contains lines of oc command to execute. Don't use command that could return '1' as expected.
      - action: <path to file>
        resource: customize
        persona: admin
        concurrency: 1
```
example of the action file
```
oc get clusterversion
oc get co | awk '$3=="False" || $4=='True' || $5=="True"'
oc get pods -A | awk '$4!="Running" && $4!="Completed" && $4!="STATUS"'
```

**Exit Criteria**
User can customize any oc commands and run concurrently with multiple users.

### DIFF
--- a/reliability/tasks/CustomizedTask.py
+++ b/reliability/tasks/CustomizedTask.py
@@ -1,0 +1,66 @@
+from .GlobalData import global_data
+from tasks.utils.oc import oc
+import os
+import logging
+
+class OCTask():
+    def __init__(self, command, kubeconfig):
+        self.command = command
+        self.kubeconfig = kubeconfig
+        
+    def execute(self):
+        return oc(self.command.lstrip("oc "), self.kubeconfig)
+
+class FileTask():
+    def __init__(self, file, kubeconfig):
+        self.file =file
+        self.kubeconfig = kubeconfig
+        self.output = []
+        self.rc = 0
+    def execute(self):
+        with open (self.file) as f:
+            for l in f.readlines():
+                task = OCTask(l, self.kubeconfig)
+                output, rc = task.execute()
+                self.output.append(output)
+                if rc != 0:
+                    break
+        return self.output, rc
+
+class CustomizedTask():
+    def __init__(self):
+        self.output = None
+        self.code = 0
+        self.customized_task_succeeded = 0
+        self.customized_task_failed = 0
+        self.logger = logging.getLogger('reliability')
+
+    def execute_task(self, action, kubeconfig):
+        if action.startswith("oc ", 0):
+            task = OCTask(action, kubeconfig)
+        elif os.path.isfile(action):
+            task = FileTask(action, kubeconfig)
+        else:
+            self.logger.warning(f"Customized action '{action}' is not supported. Provide an oc command or a file contains lines of oc command.")
+            return None
+        outout, rc = task.execute()
+        with global_data.customized_task_lock:
+            if rc == 0:
+                self.customized_task_succeeded += 1
+                return "Customized action succeeded for action: " + action
+            else:
+                self.customized_task_failed += 1
+                return "Customized action failed for action: " + action
+
+customizedTask = CustomizedTask()
+
+
+if __name__ == "__main__":
+    kubeconfig = "<path to kubeconfig of the cluster>"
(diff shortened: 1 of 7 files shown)


qiliRedHat:

paigerube14:
LGTM, tested with action in the config and in a file, both worked as expected

reliability/tasks/CustomizedTask.py
```diff
@@ -0,0 +1,65 @@
...
+                return "Customized action succeeded for action: " + action
+            else:
+                self.customized_task_failed += 1
+                return "Customized action failed for action: " + action
+
+customizedTask = CustomizedTask()
+
+
+if __name__ == "__main__":
+    kubeconfig = "/Users/qili/git/svt/reliability/kubeconfigs/kubeconfig_kubeadmin"
```
kedark3, line 59: should user be allowed to change this path? Also seems like `qili` is your private filepath, I feel like that shouldn't be in the codebase.

kedark3 (approved):
Looks nice :100:  I hope you don't mind me just peeking and leaving a review!1

reliability/tasks/CustomizedTask.py
qiliRedHat: @kedark3 Hi KK, thank you very much for reviewing my code! Good question. The main here is for test purpose of this single python - as we con't have UT code - should add when time is right. 'Real' run will never hit this main. But you're quite right, I'll replace the personal path to a common placeholder.

kedark3: Right, I understand that aspect of it :) But as you said, still good to replace it with a placeholder.

qiliRedHat: https://github.com/openshift/svt/pull/676/commits/a3f6255c38c847c238463fbdf0344cc05448af61 did the change. I'll squash the commits later.

qiliRedHat: Squashed

reliability/tasks/CustomizedTask.py
```diff
@@ -0,0 +1,66 @@
...
+import os
+import logging
+
+class OCTask():
+    def __init__(self, command, kubeconfig):
+        self.command = command
+        self.kubeconfig = kubeconfig
+        
+    def execute(self):
+        return oc(self.command.lstrip("oc "), self.kubeconfig)
```
mffiedler, line 12: This is fine for now, but a future enhancement could be to run arbitrary external commands.   This might be useful for monitoring system resources or triggering actions outside of the constraints of "oc"

reliability/tasks/Users.py
```diff
@@ -46,6 +46,6 @@ def init(self):
     
 if __name__ == "__main__":
     all_users.init()
-    all_users.load_users("/home/mifiedle/mffiedler_git/svt/reliability/config/users.spec")
```
mffiedler, line 49: Oooops.   Thanks.

mffiedler (approved):
LGTM

reliability/tasks/CustomizedTask.py
qiliRedHat: Hi Mike, Thanks for the comment. I will consider this in future enhancement.

PR #682, opened 2021-08-20
Changing image path to quay.io

skordas:
Update image path to be sure it's always downloaded from quay.io
Increasing number of running cases


paigerube14 (approved):
LGTM

PR #699, opened 2021-11-18
Adding upgrade script and helper functions for loaded-upgrades in ocp…

paigerube14:
This is an upgrade script and write to file scripts for loaded-upgrades. Not sure if here is the right spot for this. Can move into the ocp-qe-perfscale-ci repo. Thoughts?


kedark3:
@paigerube14 can we colocate these scripts in the same branch as the upgrade job in the ocp-qe-scale-ci ? And reference directly from there?

paigerube14:
> @paigerube14 can we colocate these scripts in the same branch as the upgrade job in the ocp-qe-scale-ci ? And reference directly from there?

Sure, definitely think that's a better location, will move and update necessary scripts

paigerube14:
Moving to ocp repo, closing this

PR #689, opened 2021-09-14
Large Number of Namespaces with Network policy 

paigerube14:
Added script with steps outlined in the test case below

Automating test case: https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-26279

PR #702, opened 2022-02-07
Taking out OVN option

paigerube14:
Qiujie reminded me that in the [TC](https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-26279) it specifically says this test case only works for sdn. I did try on OVN as well and it doesnt get the same rules in any of the containers

### DIFF
--- a/networking/synthetic/scripts/large_network_policy.sh
+++ b/networking/synthetic/scripts/large_network_policy.sh
@@ -22,12 +22,6 @@ done
 
 network_namespace="openshift-sdn"
 container_name="sdn"
-sdn_project_count=$(oc get projects | grep sdn | wc -l | xargs)
-echo "sdn count $sdn_project_count"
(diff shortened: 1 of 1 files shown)


qiliRedHat:
LGTM. Is OVN not supported? Why @mffiedler said 'We should run this test on openshift-sdn and ovn-kubernetes' in https://github.com/openshift/svt/pull/689#pullrequestreview-790515268.

mffiedler:
@paigerube14 @qiliRedHat  Looking at old PRs - do we still want this one?   Sorry for delay.

qiliRedHat:
I don't understand why network policy is not applied on OVN. But Paige did mentioned 'I did try on OVN as well and it doesnt get the same rules in any of the containers'. I think we can do more investigation while running this regression test during 4.11 and update here.

PR #704, opened 2022-02-28
Adding large scale concurrent builds

paigerube14:
Very similar to concurrent builds, just using an online nodejs mongo template from [here](https://github.com/openshift/online/tree/master/ansible/roles/oso_template_imagestream_sync/files/starter/templates) 

Using yq to do some yaml parsing, so adding to the cluster loader parameters

### DIFF
--- a/openshift_performance/ci/content/conc_builds_nodejs_mongo.yaml
+++ b/openshift_performance/ci/content/conc_builds_nodejs_mongo.yaml
@@ -1,0 +1,17 @@
+projects:
+  - num: 500
+    basename: svt-nodejs-
+    ifexists: default
+    tuning: default
+
+    templates:
+      - num: 1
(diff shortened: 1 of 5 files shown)


qiliRedHat:
@paigerube14 I saw you already added concurrent builds into scale-ci and there is nodejs mongodb APP covered https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/kube-burner/builds/nodejs.sh. Are there something that cannot be covered by the scale-ci and need to be run with the scripts(I guess there should be related polarion regression test case) here?

qiliRedHat:
@paigerube14 Thanks for the clarification! As this pr had been made long time ago, let's re-test it to make sure things are still working. Think we can run the test in coming regression test run.

PR #709, opened 2022-04-06
add slack notifications when visiting app fails in new_app and load_a…

qiliRedHat:
…pp task
https://issues.redhat.com/browse/OCPQE-8913

In the current reliability test, when there is a failure, slack channel will get notification. This is to reduce the effort of monitoring the long run reliability test.
Currently there when there is 'load app' failure, reliability test doesn't slack notification, as there could be some flaky of application access error, sending notification could be a noise to the channel.
But in some rare case, the application isn't up and running due to problem such as template issues, the test is failing but tester doesn't get notification until checking log file.
I want to add notification when application visiting fails, at the same time avoid too much noise of flaky app load failure.
As now https://issues.redhat.com/browse/OCPQE-9017 added the ability to avoid slack message flood. It's possible to add notification when application visiting fails.

### DIFF
--- a/reliability-v2/integrations/SlackIntegration.py
+++ b/reliability-v2/integrations/SlackIntegration.py
@@ -19,7 +19,8 @@ class SlackIntegration:
         self.limited_messages = ("Unable to connect to the server: dial tcp",
         "Unable to connect to the server: dial tcp: lookup",
         "error: You must be logged in to the server (Unauthorized)",
-        "Result: Error from server (AlreadyExists): project.project.openshift.io")
+        "Result: Error from server (AlreadyExists): project.project.openshift.io",
+        "load_app: visit route")
     
     def init_slack_client(self, slack_channel, slack_member):
         # Init slack client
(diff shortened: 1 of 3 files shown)


qiliRedHat:
Tests:
new_app failed notification when app access not succeeded after retry times.
[2022-04-06 16:08:34 CST] @qili,  :boom: new_app: visit 'django-psql-persistent-testuser-0-0.apps.qili-preserve-aws-lr.qe-lrc.devcluster.openshift.com' failed after 60 retries. status_code: 503

load_app failed notification when get route failed.
[2022-04-06 16:46:45 CST] @qili,  :boom: load_app: get route failed for namespace 'testuser-0-0'.

load_app failed notification when some of the app accesses return non 200.
[2022-04-06 17:04:30 CST] @qili,  :boom: load_app: visit route 'nodejs-postgresql-persistent-testuser-0-0.apps.qili-preserve-aws-lr.qe-lrc.devcluster.openshift.com' failed. status_code: 503

qiliRedHat:
@mffiedler @paigerube14 PTAL

PR #716, opened 2022-07-11
Template apiversion update

qiliRedHat:
apiVersion of kind "Template" were changed from "v1" to "template.openshift.io/v1" 
Update all related files in svt repo.


qiliRedHat:
@paigerube14 @mffiedler PTAL This caused problem in 2 of regression test cases I attempted for 4.11. I updated all related json files in this repo.

paigerube14 (approved):
I hit this as well, looks good to me!

PR #734, opened 2022-09-28
Start to loaded-projects test in kube-burner

paigerube14:
Adding loaded projects workloads to use kube-burner as a custom workload to load the cluster

TC: https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=&revision=10601452
Work jira card: https://issues.redhat.com/browse/OCPQE-12260


svetsa-rh:
@paigerube14, do you have a passing Jenkins run with this automated test case that I can look at?

paigerube14:
@svetsa-rh yes sorry linked it in the regression-test job pr but not here: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/regression-test/6/console

qiliRedHat:
@paigerube14 code looks good to me. But when I read the polarion test case, I saw with 3 workers, `Rerun script with 99 projects`. So now our config is `## Cluster config: 3 master (m5.2xlarge or equivalent) with 3 worker`, Do you want to set the default parameter directly to 99 or you prefer to overwrite it as parameter?

Another thought is checking cpu/memory could be a common functionality in some test cases, later we can think and add this, so test case like this can call that function to do a performance check. I opened https://issues.redhat.com/browse/OCPQE-12300 to track.

paigerube14:
> @paigerube14 code looks good to me. But when I read the polarion test case, I saw with 3 workers, `Rerun script with 99 projects`. So now our config is `## Cluster config: 3 master (m5.2xlarge or equivalent) with 3 worker`, Do you want to set the default parameter directly to 99 or you prefer to overwrite it as parameter?
> 
> 
> 
> Another thought is checking cpu/memory could be a common functionality in some test cases, later we can think and add this, so test case like this can call that function to do a performance check. I opened https://issues.redhat.com/browse/OCPQE-12300 to track.

@svetsa-rh is doing something very similar to this in her PR currently. Once hers gets merged we can reuse the same functionality to all test cases to resolve this issue

paigerube14:
@qiliRedHat @svetsa-rh since we are using a custom kube-burner workload type we can use the benchmark-compare code along with the metrics.yaml collection that's already in place to look at the CPU and memory of the nodes

https://github.com/cloud-bulldozer/e2e-benchmarking/blob/ec219202f02e766214b9c66618b41c717a58189a/workloads/kube-burner/metrics-profiles/metrics.yaml#L47

I think we need to better align the regression-test and kube-burner to set the comparison_config, email_id and google sheet for output

qiliRedHat:
@paigerube14 @svetsa-rh Thanks for the ongoing work and the comments! let's track the progress and add discussions in a Jira, if not already created already by Sharada, let's use https://issues.redhat.com/browse/OCPQE-12300.

PR #733, opened 2022-09-27
Regr automation migration and enhancement for replace loaded worker nodes with larger instance types.

svetsa-rh:
See [OCPQE-11961](https://issues.redhat.com//browse/OCPQE-11961) and [OCPQE-12152](https://issues.redhat.com//browse/OCPQE-12152)

https://issues.redhat.com/browse/OCPQE-11961
https://issues.redhat.com/browse/OCPQE-12152

Test case: https://polarion.engineering.redhat.com/polarion/redirect/project/OSE/workitem?id=OCP-40557

Also added parameters to specify cloud type and number of replicas to create when creating new machineset.

Here are couple of successful runs:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/regression-test/67/console
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/regression-test/65/console

@paigerube14 and @qiliRedHat PTAL

PR #743, opened 2022-11-15
Large network policy update

qiliRedHat:
1. Remove extra env file from the common [custom_workload_env.sh file](https://github.com/openshift/svt/blob/e213e7b99803d410e00cb9c56172905607868320/perfscale_regression_ci/scripts/custom_workload_env.sh)
2. A little refine to get sdn pods

@svetsa-rh 
I saw you added these [2 lines](https://github.com/openshift/svt/blob/e213e7b99803d410e00cb9c56172905607868320/perfscale_regression_ci/scripts/custom_workload_env.sh#L9-L10) in [PR](https://github.com/openshift/svt/pull/733) . I found the 2 lines doesn’t work in my test case [large_network_policy.sh](https://github.com/openshift/svt/blob/master/perfscale_regression_ci/scripts/network/large_network_policy.sh).  I know you added them to get the performance of worker nodes for your enhanced test case. I think you should be using some extra metrics files too to get those metrics into es, is that this [line](https://github.com/svetsa-rh/svt/blob/e08813cf8dfcc273665fb0cba2268a8bc4c263ab/perfscale_regression_ci/scripts/scalability/replace_loadedworker_nodes_env.sh#L6)?
```
11-15 02:48:44.955  2022-11-14 18:48:44,898 - touchstone - ERROR - Error: Issue capturing results from elasticsearch using config {'filter': {'metricName.keyword': 'nodeCPU-Masters'}, 'buckets': ['labels.mode.keyword', 'labels.instance.keyword'], 'aggregations': {'value': ['max']}}
11-15 02:48:45.586  2022-11-14 18:48:45,215 - touchstone - ERROR - Error: Issue capturing results from elasticsearch using config {'filter': {'metricName.keyword': 'nodeMemoryUtilization-Masters'}, 'buckets': ['labels.instance.keyword'], 'aggregations': {'value': ['max']}}
```

I suggest we can keep these 2 lines only in your test case’s specific env file. 
```
export COMPARISON_CONFIG="clusterVersion.json nodeMasters-max.json nodeAggWorkers.json"
export GEN_CSV=true
```

Please let me know if this change works for your test cases.

### DIFF
--- a/perfscale_regression_ci/scripts/custom_workload_env.sh
+++ b/perfscale_regression_ci/scripts/custom_workload_env.sh
@@ -6,5 +6,5 @@ export WORKLOAD=${WORKLOAD:-"custom"}
 export INDEXING=${INDEXING:-"false"}
 export METRICS_PROFILE=${METRICS_PROFILE:-"metrics-profiles/metrics.yaml"}
 export ALERTS_PROFILE=${ALERTS_PROFILE:-""}
-export COMPARISON_CONFIG="clusterVersion.json nodeMasters-max.json nodeAggWorkers.json"
-export GEN_CSV=true
+export COMPARISON_CONFIG="clusterVersion.json"
+export GEN_CSV=false
--- a/perfscale_regression_ci/scripts/network/large_network_policy.sh
+++ b/perfscale_regression_ci/scripts/network/large_network_policy.sh
@@ -40,7 +40,7 @@ sdn_project_count=$(oc get projects | grep sdn | wc -l | xargs)
 #   container_name="ovnkube-node"
 # fi
 
-network_pods=$(oc get pods -n $network_namespace -o name)
+network_pods=$(oc get pods -n $network_namespace -o name | grep -v controller)
 
 echo $network_pods
 


qiliRedHat:
@svetsa-rh @paigerube14 PTAL

paigerube14:
@qiliRedHat I think for environment variables that don't work for your test that you should add in your own enviornment variables to set the values that you want and leave the ones in the [custom_workload_env.sh file](https://github.com/openshift/svt/blob/e213e7b99803d410e00cb9c56172905607868320/perfscale_regression_ci/scripts/custom_workload_env.sh)

svetsa-rh:
Hi @qiliRedHat I think you just need to add below to custom_workload_env.sh to overcome those errors
export METRICS_PROFILE=${METRICS_PROFILE:-"metrics-profiles/metrics.yaml"}

qiliRedHat:
> @qiliRedHat I think for environment variables that don't work for your test that you should add in your own enviornment variables to set the values that you want and leave the ones in the [custom_workload_env.sh file](https://github.com/openshift/svt/blob/e213e7b99803d410e00cb9c56172905607868320/perfscale_regression_ci/scripts/custom_workload_env.sh) Thoughts?

@paigerube14 Do you mean you want to keep or not keep the following lines in /perfscale_regression_ci/scripts/custom_workload_env.sh?
```
#export COMPARISON_CONFIG="clusterVersion.json nodeMasters-max.json nodeAggWorkers.json"
#export GEN_CSV=true
```

My thinking is custom_workload_env.sh is the most basic one, all test cases including that don't have their own environment variables should work with it. So we should keep it to have the most basic variables that can work (don't break) any test.

qiliRedHat:
> Hi @qiliRedHat We need to update METRICS_PROFILE in custom_workload_env.sh to use metrics-aggregated.yaml instead export METRICS_PROFILE=${METRICS_PROFILE:-"metrics-profiles/metrics-aggregated.yaml"}
> 
> metrics-aggregated.yaml actually is the right one to use. kube burner sets it by default, so my tests didn't have issue and I missed noticing it.

@svetsa-rh The workload we run is custom workload, it doesn't any default metrics. Metrics can be defined by user depending one what metrics the user want to collect. Adding METRICS_PROFILE=${METRICS_PROFILE:-"metrics-profiles/metrics-aggregated.yaml"} can make it work with COMPARISON_CONFIG="clusterVersion.json nodeMasters-max.json nodeAggWorkers.json", but I'm afraid if some user defines other METRICS_PROFILE in their test case's env file that is not compatible with the 'default' COMPARISON_CONFIG and visa versa.

svetsa-rh:
> > Hi @qiliRedHat We need to update METRICS_PROFILE in custom_workload_env.sh to use metrics-aggregated.yaml instead export METRICS_PROFILE=${METRICS_PROFILE:-"metrics-profiles/metrics-aggregated.yaml"}
> 
> > 
> 
> > metrics-aggregated.yaml actually is the right one to use. kube burner sets it by default, so my tests didn't have issue and I missed noticing it.
> 
> 
> 
> @svetsa-rh The workload we run is custom workload, it doesn't any default metrics. Metrics can be defined by user depending one what metrics the user want to collect. Adding METRICS_PROFILE=${METRICS_PROFILE:-"metrics-profiles/metrics-aggregated.yaml"} can make it work with COMPARISON_CONFIG="clusterVersion.json nodeMasters-max.json nodeAggWorkers.json", but I'm afraid if some user defines other METRICS_PROFILE in their test case's env file that is not compatible with the 'default' COMPARISON_CONFIG and visa versa.

@qiliRedHat i understand what you mean. I have no objection to go either way.

qiliRedHat:
@paigerube14 Please let me know if you have more comment on https://github.com/openshift/svt/pull/743#issuecomment-1316179637. If it looks good, please help to merge this.

perfscale_regression_ci/scripts/custom_workload_env.sh
```diff
@@ -6,5 +6,5 @@ export WORKLOAD=${WORKLOAD:-"custom"}
 export INDEXING=${INDEXING:-"false"}
 export METRICS_PROFILE=${METRICS_PROFILE:-"metrics-profiles/metrics.yaml"}
 export ALERTS_PROFILE=${ALERTS_PROFILE:-""}
-export COMPARISON_CONFIG="clusterVersion.json nodeMasters-max.json nodeAggWorkers.json"
-export GEN_CSV=true
\ No newline at end of file
+#export COMPARISON_CONFIG="clusterVersion.json nodeMasters-max.json nodeAggWorkers.json"
```
paigerube14, line 9: I think instead of commenting these out we should just have clusterversion as the default for comparison config and gen_csv false and other scripts can reset and generate csv as they want 

@svetsa-rh we found you need to update your scripts to set the metrics profile to metrics-aggregated correct? Can we open an issue  to make sure we update that properly

qiliRedHat: @paigerube14 Updated comparison config to have clusterversion and gen_csv to false.

svetsa-rh: @paigerube14 Sure, will do that.

mffiedler:
Unrelated to this specific PR:  does the large network policy script currently run on OVN?   I saw some openshift-sdn specific stuff but it wasn't clear if the script also works on OVN.  I haven't run it myself yet :-)

qiliRedHat:
@mffiedler It is run on SDN only. I've tried it on OVN but the test case can not be executed correctly. I've talked to zhanqi, he told me the below step can only work on SDN, because OVN does't have br0, andtable=80 is defined in openshift-sdn too. 
```
count=$(oc exec $pod -n $network_namespace -c $container_name -- ovs-ofctl dump-flows br0 -O openflow13 | grep table=80 | wc -l | xargs)
```

PR #744, opened 2022-11-17
add script to install ingress autoscaling

qiliRedHat:


openshift_scalability/install-ingress-autoscaling.sh
```diff
@@ -0,0 +1,173 @@
...
+    echo "[ERROR] please make sure KedaController is successfully installed."
+    exit 1
+fi
+echo "[INFO ]====Installing Ingresss Autoscaling===="
+
+echo "[INFO] Creating project openshift-ingress-operator"
+oc project openshift-ingress-operator
+
+echo "[INFO] Applying ConfigMap cluster-monitoring-config"
+oc apply -f -<<EOF
```
paigerube14, line 32: Let's add all these yaml files to the content folder or in a different folder in this directory to be reusable

qiliRedHat:
@paigerube14 PTAL

openshift_scalability/scripts/ingress_autoscaling/install-ingress-autoscaling.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+oc apply -f content/ingress-scaler.yaml
+
+echo "[INFO] ====Post Install Check===="
+echo "[INFO] Getting hpa"
+oc get hpa
+echo "[INFO] Getting ingresscontroller replicas"
+oc get ingresscontroller/default -o yaml -n openshift-ingress-operator | grep replicas:
+echo "[INFO] Getting number of worker nodes:"
+oc get nodes --no-headers -l node-role.kubernetes.io/worker= | wc -l
```
paigerube14, line 85: I think there should be some validation here? I'm not 100% sure what should be the result. I was able to properly run with the help of the comments and the validation you added above :)

qiliRedHat: This is a script to help enabling the feature before running the actual test cases. I haven't automate the test cases yet. But you remind me it's also good to add part of the validations here so I don't need to do them again on the test case side.

qiliRedHat:
@paigerube14 PTAL

qiliRedHat:
Merged after getting lgtm in peer review.

PR #745, opened 2022-11-18
add test case create-delete-pod-ensure-service

qiliRedHat:
https://issues.redhat.com/browse/OCPQE-12153


qiliRedHat:
@mffiedler @paigerube14 PTAL

paigerube14:
I'm currently not able to run this test case because of bug: https://issues.redhat.com/browse/OCPBUGS-1028 https://bugzilla.redhat.com/show_bug.cgi?id=2097830 
I did notice that even when I set the slack notifications to false it was still giving me errors that's its not able to post to slack 
`2023-01-09 17:55:02,838 - WARNING - post_error_to_slack had exception: ''NoneType' object has no attribute 'chat_postMessage'`

Think you just need to add [this same if statement ](https://github.com/openshift/svt/blob/66ebb47589200a09e036dab02830e2ec8807dbbf/reliability-v2/integrations/SlackIntegration.py#L93) to info() and error()

qiliRedHat:
> I'm currently not able to run this test case because of bug: https://issues.redhat.com/browse/OCPBUGS-1028 https://bugzilla.redhat.com/show_bug.cgi?id=2097830

@paigerube14 I had the same issue on MAC too. I run the test in jumphost. You can use mine if needed, ping me for the host info.

> Think you just need to add [this same if statement ](https://github.com/openshift/svt/blob/66ebb47589200a09e036dab02830e2ec8807dbbf/reliability-v2/integrations/SlackIntegration.py#L93)to info() and error()

@paigerube14 Fixed, thanks!
`grep 'chat_postMessage' reliability.log` no more warning is seen in log now. 
With slack enabled, works well too.

paigerube14:
Used your jump host and everything looks good!! 
/lgtm

PR #754, opened 2023-04-04
add storage to monitoring, update readme

qiliRedHat:
Changes
- persist monitoring to avoid data lost during upgrade https://issues.redhat.com/browse/OCPQE-12771
- Some minor updates for README and add more comments to start.sh


qiliRedHat:
@svetsa-rh PTAL thanks.

reliability-v2/start.sh
```diff
@@ -217,6 +219,43 @@ elif [[ ! -z $kubeconfig && ! -z $users_spec && ! -z $kubeadmin_password ]]; the
...
+    oc rollout status -n openshift-monitoring deploy/cluster-monitoring-operator
+    oc rollout status -n openshift-monitoring sts/prometheus-k8s
+    token=$(oc create token -n openshift-monitoring prometheus-k8s --duration=6h)
+    URL=https://$(oc get route -n openshift-monitoring prometheus-k8s -o jsonpath="{.spec.host}")
+    prom_status="not_started"
+    sleep 30
+    while [[ "$prom_status" != "success" ]]; do
+        prom_status=$(curl -s -g -k -X GET -H "Authorization: Bearer $token" -H 'Accept: application/json' -H 'Content-Type: application/json' "$URL/api/v1/query?query=up" | jq -r '.status')
+        sleep 5
+        echo "Prometheus status not ready yet, retrying in 5s..."
```
svetsa-rh, line 253: @qiliRedHat In case prometheus is taking forever to be ready - instead of infinite loop until prometheus is ready, how about a timeout and gracefully exiting with the reason for the failure why prometheus is not ready?

svetsa-rh: Looks good to me otherwise!

qiliRedHat: @svetsa-rh Thank you!
I added what you suggested.
Test result:
```
Sleep 30s to wait for prometheus status to become success.
Prometheus status is not success yet, retrying in 10s, retries left: 19.
Prometheus is success now.
```
Yes this is similar to post config, but does't use INFRA node - to save resource in a long run. Thanks for the info about that issue. I haven't see this in my recent runs with this branch yet. I'll keep an eye on it.

qiliRedHat: @svetsa-rh PTAL again, thanks

PR #763, opened 2023-07-14
test for large object list

lenahorsley:
Test output can be found in the comment section of https://issues.redhat.com/browse/OCPQE-15073

@paigerube14 
@qiliRedHat

### DIFF
--- a/openshift_performance/ci/scripts/large_object_list/large_object_list.sh
+++ b/openshift_performance/ci/scripts/large_object_list/large_object_list.sh
@@ -12,15 +12,13 @@
 
 #! /bin/bash
 
-
-
-expected_namespaces=250
-expected_secrets=5000
-secrets_per_namespace=20
+expected_namespaces=${1:-100}
+secrets_per_namespace=${2:-10}
+expected_secrets=$(( $expected_namespaces * $secrets_per_namespace ))
+debug_secrets_command="oc get secrets -A -l test=listobject --no-headers -v9"
 my_secret_name="-secret-"
 my_namespace="objecttest-"
 my_object_type="namespaces"
-
 
 # simple function to display the status of operators and nodes
 function get_operator_and_node_status() {
@@ -83,9 +81,13 @@ function delete_namespaces() {
 echo "Checking the cluster for test objects..."
 check_cluster=$(oc get $my_object_type -A| grep $my_namespace | wc -l)
 if [ $check_cluster -ne 0 ]; then
-   echo "Removing test objects from the cluster before test execution"
-   delete_namespaces $check_cluster
+  echo "Removing test objects from the cluster before test execution"
+  delete_namespaces $check_cluster
+else
+  echo "No test objects present..."
 fi
+
+echo ""
 
 # Step 2. Start the log file
 echo -e "\nWriting subprocess logs to ./object-list-test.log"
@@ -102,7 +104,6 @@ echo ""
 echo "$(date) - Namespace and secret creation starting..."
 cycle_start_time=`date +%s`
 create_namespaces_and_secrets
-sleep 5
 
 echo "$(date) - Namespace and secret creation complete."
 echo -e "\nNamespace and secret creation complete." >> object-list-test.log &
@@ -112,16 +113,12 @@ echo "Total time for create cycle: $total_cycle_time s."
 
 
 # Pause to allow the system to catch up
-echo -e "\nSleeping for 5 minutes..."
-echo -e "\nSleeping for 5 minutes..." >> object-list-test.log &
+echo -e "\nSleeping for 5 minutes..." | tee object-list-test.log
 sleep 5m
 
-echo -e "\nChecking for the namespaces and secrets."
-echo -e "\nChecking for the namespaces and secrets." >> object-list-test.log &
+echo -e "\nChecking for the namespaces and secrets." | tee object-list-test.log
 
 
-# Step 4. Count the secrets and namespaces, then compare the actual and expected values.
-# Find the secrets created by this test and print the output to the log file
 my_namespace_command='oc get secrets -A -l test=listobject &>> object-list-test.log &'
 eval $my_namespace_command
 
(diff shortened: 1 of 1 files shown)


openshift_performance/ci/scripts/large_object_list/large_object_list.sh
```diff
@@ -0,0 +1,176 @@
...
+echo "$(date) - Namespace and secret creation complete."
+echo -e "\nNamespace and secret creation complete." >> object-list-test.log &
+cycle_end_time=`date +%s` 
+total_cycle_time=$((cycle_end_time - cycle_start_time))
+echo "Total time for create cycle: $total_cycle_time s."
+
+
+# Pause to allow the system to catch up
+echo -e "\nSleeping for 5 minutes..."
+echo -e "\nSleeping for 5 minutes..." >> object-list-test.log &
```
qiliRedHat, line 116: use '| tee object-list-test.log' can output to console at the same time write to a log file

openshift_performance/ci/scripts/large_object_list/large_object_list.sh
```diff
@@ -0,0 +1,176 @@
...
+echo -e "\n===========================================" >> object-list-test.log &
+echo "$(date) - NEW RUN" >> object-list-test.log &
+echo ""
+
+
+# Step 3. Create the namespaces and the secrets. The output is printed to a log file. Also, we are calculating test object creation time.
+echo "$(date) - Namespace and secret creation starting..."
+cycle_start_time=`date +%s`
+create_namespaces_and_secrets
+sleep 5
```
qiliRedHat, line 105: Why sleep 5 here? Is that to wait the background process to be finished?

openshift_performance/ci/scripts/large_object_list/large_object_list.sh
```diff
@@ -0,0 +1,176 @@
...
+##       while true; do oc get co --no-headers| grep -v 'True.*False.*False'; oc get nodes --no-headers| grep -v ' Ready'; date; sleep 10; done
+################################################ 
+
+#! /bin/bash
+
+
+
+expected_namespaces=250
+expected_secrets=5000
+secrets_per_namespace=20
```
qiliRedHat, line 19: 1. Can we make these as parameters or overwrite by env variables? 
2. expected_secrets can be calculated the the expected_namespaces x secrets_per_namespace in code. So we only need 2 inputs.

openshift_performance/ci/scripts/large_object_list/large_object_list.sh
```diff
@@ -0,0 +1,176 @@
...
+else
+    echo "Test failed"
+    echo "Operators (oc get co --no-headers| grep -v 'True.*False.*False'):"
+    oc get co --no-headers| grep -v 'True.*False.*False'
+    echo ""
+    echo "Nodes oc get nodes --no-headers | grep -v Ready):"
+    oc get nodes --no-headers | grep -v Ready
+    echo ""
+    echo "expected # of namespaces: $expected_namespaces.  actual # of namespaces: $actual_namespaces"
+    echo "expected # of secrets: $expected_secrets.  actual # of secrets: $actual_secrets"
```
qiliRedHat, line 173: If the number is not as expected, we want to check the exact output of the get command, better to with ' -v9' to get debug info
oc get secrets -A -l test=listobject --no-headers -v9

qiliRedHat: And we can point to the output file location in the result section

openshift_performance/ci/scripts/large_object_list/large_object_list.sh
```diff
@@ -0,0 +1,177 @@
...
+  echo ""
+  oc get co
+  echo ""
+}
+
+# function to create namespaces and secrets
+function create_namespaces_and_secrets() {
+
+  for ((i=0; i<$expected_namespaces; i++)); do
+      my_namespace_command='oc new-project --skip-config-write "${my_namespace}${i}" > /dev/null && echo -e "\n\nCreated project ${my_namespace}${i}" &>> object-list-test.log &'
```
paigerube14, line 37: Lets try to use kube-burner or the cluster-loader script to create the objects here 

Think you can use kube-burner and use some of the common functions if you moved this to the [perfscale_regression_ci folder](https://github.com/openshift/svt/tree/master/perfscale_regression_ci/scripts/performance)

Could set _replicas_ = _secrets_per_namespace_ in the config file and _JOB_ITERATION_ = _expected_namespaces_ 

Example config that'll be very similar to what you need: https://github.com/openshift/svt/blob/master/perfscale_regression_ci/kubeburner-object-templates/blank-projects-config.yml

lenahorsley: In the Jira task, Qiujie stated the script should be done in a way a customer can run the test. @qiliRedHat What's your opinion?

paigerube14: Customers should be able to run kube-burner since it is open source. 

How long does the creation of the objects take? With kube-burner it should be quicker and will help verify the objects created

I'm not super tied to how to run this, just want to stay up to date with the tooling we are using other places

openshift_performance/ci/scripts/large_object_list/large_object_list.sh
```diff
@@ -0,0 +1,177 @@
...
+      for ((j=0; j<$secrets_per_namespace; j++)); do
+         echo "
+         apiVersion: v1
+         kind: Secret
+         metadata:
+           labels:
+             test: listobject
+           namespace: ${my_namespace}${i}
+           name: ${my_namespace}${i}${my_secret_name}${j}
+         " | oc apply -f - &>> object-list-test.log &
```
qiliRedHat, line 48: There is no rate control here. I remember in the bad actor case, we have rate control, if parallel rate is higher than some number, there will be some issue.

paigerube14: if this test case doesn't require parallel object creation we should use kube-burner. looking at the code it doesnt appear to create kube-burner
I think we only kept in other place because it required creating objects in parallel that kube-burner doesnt support

lenahorsley:
Closing this PR. I'll open a new one with the suggested changes.

qiliRedHat:
@paigerube14 I realized this PR isn't merged when I tried to run https://polarion.engineering.redhat.com/polarion/#/project/OSE/workitem?id=OCP-65206. So I reopened it.

qiliRedHat:
I found https://github.com/openshift/svt/pull/770 is the new version of this change. Close it.

PR #768, opened 2023-09-11
taking out extra logging lines

paigerube14:
Taking out a few of the logging lines that may be overkill


svetsa-rh:
@paigerube14 Change looks good to me, but should we keep it as logger.debug at all or no? Will there be any useful info we will miss if running on Jenkins?

paigerube14:
No, these really just cause noise and don't help with the actual end output or debugging

PR #769, opened 2023-09-12
Fix rosa admin cmd

qiliRedHat:
Fix rosa admin cmd
Avoid warning in check_replicas


qiliRedHat:
@liqcui PTAL. I made some fixes.

PR #782, opened 2024-01-17
add reliability config templates

qiliRedHat:
add multiple reliability config template for different test clusters


qiliRedHat:
@liqcui PTAL
Tested -c with a new run https://redhat-internal.slack.com/archives/C0266JJ4XM5/p1705467033125269

PR #798, opened 2024-10-29
integrate netobserv to reliability tests

memodi:
[NETOBSERV-1874](https://issues.redhat.com//browse/NETOBSERV-1874) integrate netobserv to reliability tests
dependent on https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/659 

Additionally, it adds new improvements and fixes some bugs:
- add trap/cleanup code.
- convert dittybopper code into a function and fix a bug to check for performance-dashboard directory.
- import dashboards.

### DIFF
--- a/reliability-v2/config/reliability-netobserv.yaml
+++ b/reliability-v2/config/reliability-netobserv.yaml
@@ -14,7 +14,7 @@ reliability:
       # For rosa cluster created in Prow, the admin user_name is rosa-admin.
       user_name: <admin_username> # admin username as kubeadmin or rosa-admin.
       loops: forever # run group for loops times. integer > 0 or 'forever', default is 1.
-      trigger: 600 # wait trigger seconds between each loop
+      trigger: 300 # wait trigger seconds between each loop
       jitter: 0 # randomly start the users in this group in trigger seconds. Default is 0.
       interval: 10 # wait interval seconds between tasks.
       # If admin network policy function is planed in the test, uncomment the following line
@@ -24,7 +24,9 @@ reliability:
         - func apply_nonamespace -p "<path_to_content>/network-policy/02_anp_allow-traffic-egress-cidr-cluster-network-p20.yaml"
         - func apply_nonamespace -p "<path_to_content>/network-policy/04_anp_allow-traffic-dev-test-p30.yaml"
         - func apply_nonamespace -p "<path_to_content>/network-policy/03_anp_allow-traffic-dev-prod-p31.yaml"
-      tasks: 
+      tasks:
+        - func check_netobserv_pods
+        - func check_flowcollector
         - func check_operators
         - oc get project -l purpose=reliability
         - func check_nodes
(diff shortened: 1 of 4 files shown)


reliability-v2/config/reliability-netobserv.yaml
```diff
@@ -57,6 +59,8 @@ reliability:
         - func check_pods -n 2 # check pods in 2 namespaces 
         - func delete_project -n 2 # delete project in 2 namespaces
         - func verify_project_deletion -n 2 # verfy project deletion in 2 namespaces
+        - func check_flowcollector
+        - func check_netobserv_pods
```
qiliRedHat, line 63: The 2 check functions should be put in only admin group, in which the tasks will be executed with kubeadmin user. The other groups users(testuser-xx) are idp users, they only have access to the project/ns that are created by themself, e.g. the test namespaces they created during the test.

reliability-v2/tasks/Tasks.py
```diff
@@ -399,3 +399,42 @@ def shell_task(self,task,user,group_name):
...
+                self.__get_kubeconfig(user),
+                ignore_log=True,
+                ignore_slack=True,
+            )
+            if rc == 0:
+                self.logger.error(f"Some pods are not Ready in ${ns} ns: {result}")
+                slackIntegration.error(f"Some pods are not Ready in ${ns}: {result}")
+                rc_return = 1
+            elif rc == 1 and result == "":
+                self.logger.info(f"Pods in ${ns} are healthy.")
```
qiliRedHat, line 438:
```suggestion
                self.logger.info(f"Pods in ns ${ns} are healthy.")
```

reliability-v2/tasks/Tasks.py
```diff
@@ -399,3 +399,42 @@ def shell_task(self,task,user,group_name):
...
+        for ns in ("netobserv", "netobserv-privileged"):
+            (result, rc) = oc(
+                f"get pods -n ${ns} -o wide --no-headers| grep -v ' Ready'",
+                self.__get_kubeconfig(user),
+                ignore_log=True,
+                ignore_slack=True,
+            )
+            if rc == 0:
+                self.logger.error(f"Some pods are not Ready in ${ns} ns: {result}")
+                slackIntegration.error(f"Some pods are not Ready in ${ns}: {result}")
```
qiliRedHat, line 435:
```suggestion
                slackIntegration.error(f"Some pods are not Ready in ns ${ns}: {result}")
```

qiliRedHat:
I can add lgtm after https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/659 is merged and integration test is done

reliability-v2/config/reliability-netobserv.yaml
memodi, line 1: @qiliRedHat do you want this to be separate config? I was thinking to have these tasks under standard reliability tasks and we'd have these checks run always. 
Of course that also means we'd have to change netobserv operator would be installed by default for all reliability 
runs and optionally exclude it, current implementation in start.sh is inverse, wdyt? this would be the initial idea had discuss so that netobserv could piggyback off your standard reliability runs.

If we have netobserv install by default, we'd need to up standard instance types or somehow have configuration that would accommodate Loki resource requirements, couple of questions here:
- Do you typically have infra nodes set up for reliability runs? if so, what are instance types of infra nodes? I can look if you can bring up that environment and I can see if Loki is able to fit on infra nodes. 
- would you be willing to up the reliability run instance types to m5.2xlarge or m5.4xlarge?

Thanks!

qiliRedHat: @memodi Reliability test has many test profiles, consider of the cost, I don't want to have noo on all profiles by default. Usually there is no infra nodes in reliability test. But reliability has the option to configure it with -i in start.sh, default size should be same as the worker, size can be configured.
If test it once (7 days) a release, I think we can do m5.2xlarge on one of the profiles.

reliability-v2/tasks/Tasks.py
```diff
@@ -399,3 +399,42 @@ def shell_task(self,task,user,group_name):
...
+                ignore_log=False,
+                ignore_slack=True,
+            )
+            if rc == 0:
+                self.logger.error(f"Some pods are not Ready in ${ns} ns: {result}")
+                slackIntegration.error(f"Some pods are not Ready in ns ${ns}: {result}")
+                rc_return = 1
+            elif rc == 1 and result == "":
+                self.logger.info(f"Pods in ns ${ns} are healthy.")
+                rc_return = 0
```
qiliRedHat, line 439: Same comment as above.

qiliRedHat:
@memodi Please squash the commits and I will add lgtm.

PR #801, opened 2024-12-03
update test config and fix some issue introduced by noo pr 798

qiliRedHat:
In this PR
1. Update test config files https://issues.redhat.com/browse/OCPQE-27394
- disable dancer-mysql-persistent in all config files because it had some issue since 4.15 test
- use dev-test-build group name to replace the previous dev-test group name, add a new dev-test group to simulate users not using openshift template
- update dev-prod group to not using openshift template because the deploymentconfig is removed from rails-pgsql-persistent template, so func scale_deploymentconfig failed
2. Fixed some issue introduced by pr https://github.com/openshift/svt/pull/798
- rc_return not set in all if branches. This caused the py exit with error and the final test results are not generated.
```
  File "/root/reliability/reliability-v2-qili-418-r01/tasks/Tasks.py", line 448, in check_netobserv_pods
    return (result, rc_return)
UnboundLocalError: local variable 'rc_return' referenced before assignment
```
3. Only check the reliability.py in this current test folders to avoid the below behavior when multiple reliability tests are running on the same jumphost.
```
The process reliability.py is still running. Waiting for it to terminate...
```
4. Adjust a proper workload for NOO reliability integration config https://issues.redhat.com/browse/OCPQE-27422


qiliRedHat:
@mehabhalodiya @memodi Please have a review.

reliability-v2/config/reliability-netobserv.yaml
```diff
@@ -47,28 +54,53 @@ reliability:
       interval: 10
       tasks:
         - func delete_all_projects # clear all projects
-        - func verify_project_deletion -n 8 # verfy project deletion in 2 namespaces
-        - func new_project -n 8 # new 8 projects
+        - func verify_project_deletion -n 2 # verfy project deletion in 2 namespaces
+        - func new_project -n 2 # new 2 projects
```
memodi, line 58: I had more scale so that we could have larger load in cluster, any reason to stick to lower amount of namespaces/workloads?

reliability-v2/config/reliability-netobserv.yaml
```diff
@@ -78,10 +110,10 @@ reliability:
         - func apply -n 8 -p "<path_to_content>/http-service.yaml"
         - func apply -n 8 -p "<path_to_content>/http-route.yaml"
       tasks:
-        - func load_app -n 8 -p 50 
-        - func scale_deployment -n 8 -p 8 # scale deployment in 2 namespaces to 2 replicas
-        - func scale_deployment -n 8 -p 5 # scale deployment in 2 namespaces to 1 replicas
-      post_tasks: 
+        - func load_app -n 8 -p 50
+        - func scale_deployment -n 8 -p 2 # scale deployment in 2 namespaces to 2 replicas
+        - func scale_deployment -n 8 -p 1 # scale deployment in 2 namespaces to 1 replicas
```
memodi, line 115: same comment about amount of load.

reliability-v2/start.sh
memodi, line 1: new path to qe-index.yaml needs to be updated at L383

memodi (changes requested):

reliability-v2/config/reliability-netobserv.yaml
qiliRedHat: @memodi As we talked in [slack](https://redhat-internal.slack.com/archives/D07R81A5G4R/p1732250299843239), the origin load is too big that exceeded the default 250 pod per node on the 3 worker nodes. See an example here: https://redhat-internal.slack.com/archives/C0266JJ4XM5/p1732269147674219?thread_ts=1732263213.922569&cid=C0266JJ4XM5
So I tried to reduce the load to see if it can work and also fit your load requirement.

reliability-v2/config/reliability-netobserv.yaml
qiliRedHat: @memodi Scaling too big here will cause a peak number of pods, as we have the 250 pod per node by default, if the peak is high, we have to reduce the base load, thus decrease the total workload. So this is a trade off, being more mild at scale can allow us to have more users/namespaces. With this new version of the load, I hope to firstly make it work (no pending pod), then we can adjust the balance between base load v.s. peak load according to your test request.

reliability-v2/start.sh
qiliRedHat: Fixed.

qiliRedHat:
@mehabhalodiya @memodi Please review again. If you have no further comment, I can get this merged.

mehabhalodiya (approved):
/lgtm

PR #807, opened 2025-01-23
NETOBSERV-2073: Check pod restarts

Amoghrd:
========reliability_result========
2025-04-04 16:00:27,153 - INFO - Reliability test results:
[Function]               |     Total|    Passed|    Failed|Failure Rate|
-----------------------------------------------------------------------
[apply_nonamespace]      |         5|         5|         0|      0.0%|
[delete_all_projects]    |        45|        45|         0|      0.0%|
[verify_project_deletion]|       202|       202|         0|      0.0%|
[new_project]            |       125|       125|         0|      0.0%|
[check_all_projects]     |        38|        38|         0|      0.0%|
[check_netobserv_pods]   |         4|         1|         3|     75.0%|
[apply]                  |       306|       306|         0|      0.0%|
[load_app]               |      1160|      1160|         0|      0.0%|
[new_app]                |        22|        22|         0|      0.0%|
[check_pods]             |        76|        76|         0|      0.0%|
[delete_project]         |       129|       125|         4|      3.1%|
[build]                  |        11|        11|         0|      0.0%|
[check_flowcollector]    |         3|         0|         3|    100.0%|
[check_operators]        |         3|         3|         0|      0.0%|
[oc_task]                |         3|         3|         0|      0.0%|
[check_netobserv_pod_restarts]|         3|         3|         0|      0.0%|
[kubectl_task]           |         3|         3|         0|      0.0%|
[cronjob.sh -n 10]       |         1|         1|         0|      0.0%|
[cronjob.sh -c]          |         1|         1|         0|      0.0%|
[scale_deployment]       |        16|        16|         0|      0.0%|
[cronjob.sh -d]          |         1|         1|         0|      0.0%|
-----------------------------------------------------------------------
========reliability_failure========
Tasks with failure rate > 1%:
[check_netobserv_pods]Failure rate:75.0%
[delete_project]Failure rate:3.1%
[check_flowcollector]Failure rate:100.0%

### DIFF
--- a/reliability-v2/tasks/Tasks.py
+++ b/reliability-v2/tasks/Tasks.py
@@ -5,11 +5,15 @@ import requests
 import sys
 from threading import Lock
 import asyncio
+import string
+import re
 from tasks.GlobalData import global_data
 from utils.cli import oc,kubectl,shell
 from utils.LoadApp import loadApp
 from integrations.SlackIntegration import slackIntegration
 from concurrent.futures import ThreadPoolExecutor
+
+netobserv_pods = {}
 
 class Tasks:
     def __init__(self):
@@ -403,7 +407,7 @@ class Tasks:
     # check flowcollector status for netobserv
     def check_flowcollector(self, user):
         self.logger.info(f"[Task] User {user}: check flowcollector")
-        # Check if nodes are Ready
+        # Check if flowcollector is Ready
         (result, rc) = oc(
             f"get flowcollector --no-headers| grep -v 'Ready'",
             self.__get_kubeconfig(user),
@@ -420,12 +424,13 @@ class Tasks:
         else:
             self.logger.error(f"Flowcollector status fetch error: result {result}, rc - {rc}")
             rc_return = 1
+        self.__log_result(rc)
         return (result, rc_return)
 
     # check netobserv pods health
     def check_netobserv_pods(self, user):
         self.logger.info(f"[Task] User {user}: check netobserv pods")
-        # Check if nodes are Ready
+        # Check if pods are Running
         for ns in ("netobserv", "netobserv-privileged"):
             (result, rc) = oc(
                 f"get pods -n {ns} -o wide --no-headers| grep -v 'Running'",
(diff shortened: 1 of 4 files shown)


reliability-v2/tasks/Tasks.py
```diff
@@ -443,4 +445,42 @@ def check_netobserv_pods(self, user):
...
+        (result, rc) = oc(
+            "get pods -n {namespace} -o jsonpath='{.items[*].metadata.name}'",
+            self.__get_kubeconfig(user),
+            ignore_log=True,
+            ignore_slack=True,
+        )
+        if rc == 0 and result != "":
+            self.logger.info(f"Pods in ns {namespace} are: {result}")  
+            pod_names = string.Split(string.Trim(result, "'"), " ")
+        return pod_names
```
qiliRedHat, line 465: Let's return the rc as well `return (pod_names, rc)`. I hope this function used directly as a task.

reliability-v2/tasks/Tasks.py
```diff
@@ -443,4 +445,42 @@ def check_netobserv_pods(self, user):
...
+    
+    # check netobserv pod restarts
+    def check_netobserv_pod_restarts(self, user):
+        self.logger.info(f"[Task] User {user}: check pod restarts")
+        # Check if pods have restarted
+        for ns in ("netobserv", "netobserv-privileged", "openshift-netobserv-operator"):
```
qiliRedHat, line 471: Another option is to make the namespaces as a string for -p parameters, like apply_nonamespace(self,user,parameter). So this function can be `def check_pod_restarts(self, user, parameter)`.
Pros: it will be more common
Cons: in logs and notifications, the task name does not include netobserv

qiliRedHat:
@Amoghrd If you have tested the updated function, would you please past a piece of test log?

Amoghrd:
@memodi Was the `check_flowcollector` and `check_netobserv_pods` passing when the code was merged?
I dont see a run in PR https://github.com/openshift/svt/pull/798 so I think it was not passing in the first place since I have not made any changes to those funcs except adding it to the final log.

Amoghrd:
The `check_netobserv_pod_restarts` func is passing though. I used the logic in pod_spy.py for this
The older funcs can be fixed in another PR.

memodi:
> @memodi Was the `check_flowcollector` and `check_netobserv_pods` passing before? I dont see a run in PR #798 so I think it was not passing then since I have not made any changes to those funcs except adding it to the final log and its failing now. Or I dont know if something has changed in between

it should be passing, except during the startup when flowcollector doesn't exist and netobserv pods are not there, but if you run longer it should be passing.

Amoghrd:
Yeah I ran it multiple times later and it seemed to give better results. Forgot to mention it here

PR #810, opened 2025-02-19
Do pip install quiet.

skordas:
This will still keep it quiet, but throw some information when something goes wrong - and it's look nicer :smile:


skordas:
@qiliRedHat  PTAL - I don't want to merge this without your check first.

qiliRedHat:
@skordas Thanks. Are you using reliabilty-v2 or play with it? I'm just curious why you had the idea to change the code here. Let me do a test to make sure it works well before I approve.

PR #838, opened 2026-04-20
reliability-v2: Fix Slack config bug and improve start.sh portability

dhensel-rh:
## Summary                                
  - Fix missing `$` on `SLACK_WEBHOOK_URL` variable check — Slack was silently
    always enabled when `SLACK_API_TOKEN` was empty because the literal string                                                                    
    "SLACK_WEBHOOK_URL" is always truthy
  - Replace fragile `sed s*'slack_enable: False'*...*` pattern with robust                                                                        
    regex-based `sed "s/slack_enable:.*/.../"` that works regardless of current value                                                             
  - Add `SLACK_API_TOKEN`, `SLACK_CHANNEL` env var injection into config, so Slack                                                                
    can be fully configured without hand-editing the YAML (previously only                                                                        
    `SLACK_MEMBER` was supported)                                                                                                                 
  - Harden haproxy version check to filter for Running pods and fall back                                                                         
    gracefully if unavailable                                                                                                                     
  - Capture `reliability.py` stdout/stderr to test folder instead of `/dev/null`                                                                  
    for debugging                                                                                                                                 
  - Support absolute paths with `-c /path/to/config.yaml` (paths starting with
    `/` are used as-is)                                                                                                                           
  - Add `${RELIABILITY_DIR}` envsubst processing so config files can reference                                                                    
    the repo root without hardcoded paths — coexists with existing `<path_to_*>`                                                                  
    placeholder system                                                                                                                            
                                                                                                                                                  
  ## Details                                                                                                                                      
  All changes are backward compatible. The default `config/reliability.yaml` is                                                                   
  NOT modified — it keeps `<path_to_*>` placeholders. The `envsubst` step only                                                                    
  substitutes `${RELIABILITY_DIR}` (restricted via single-quoted arg), leaving                                                                    
  all other variables untouched. `start.sh` already uses `envsubst` for                                                                           
  `cluster-monitoring-config.yaml` and `qe-index.yaml`, so no new dependency                                                                      
  is introduced.

### DIFF
--- a/reliability-v2/start.sh
+++ b/reliability-v2/start.sh
@@ -141,10 +141,16 @@ function generate_config {
         sed -i s*'<path_to_users_file>'*$3* $reliability_config_file
         sed -i s*'<path_to_script>'*$4* $reliability_config_file
         sed -i s*'<path_to_content>'*$5* $reliability_config_file
-        if [[ $SLACK_API_TOKEN != "" || SLACK_WEBHOOK_URL != "" ]]; then
-            sed -i s*'slack_enable: False'*'slack_enable: True'* $reliability_config_file
+        if [[ $SLACK_API_TOKEN != "" || $SLACK_WEBHOOK_URL != "" ]]; then
+            sed -i "s/slack_enable:.*/slack_enable: True/" $reliability_config_file
             if [[ $SLACK_MEMBER != "" ]]; then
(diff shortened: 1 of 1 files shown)


mehabhalodiya (approved):
LGTM! 🚀

Solid set of fixes and improvements.
All changes are backward compatible and well thought out. 
Approving. ✅

Thank you!
/lgtm

